PR #1, opened 2015-04-24
Adds a users role and adds that role to cobbler.yml and testnodes.yml

andrewschoen:
This new role will be used to create users and setup authorized_keys for
lists of users defined in group inventory files.

### DIFF
--- a/cobbler.yml
+++ b/cobbler.yml
@@ -1,5 +1,5 @@
 ---
 - hosts: cobbler 
   roles:
-    - users
+    - common
     - cobbler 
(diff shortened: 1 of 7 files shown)


andrewschoen:
@ktdreyer I default to making things roles because they're flexible. We could handle this with tags inside of the common role though.  Especially because we can give tags to a role when adding it to a playbook (which I didn't realize when I wrote this).

If you feel strongly I don't mind changing it.

andrewschoen:
@ktdreyer I lied, you actually can't provide tags to a role that *filters* the role when adding it to a playbook.  It just applies the tag you specify to everything in the role.

In light of that I think it's worth it to keep users as a role even if it never grows any bigger than it is currently.

ktdreyer:
Ok, thanks for the explanation! I'll merge this, then

PR #2, opened 2015-04-27
Add an entry for apt-mirror.front.sepia.ceph.com to /etc/hosts.

andrewschoen:
This is only a transitional thing so that one lab can talk to
apt-mirror.front.  Eventually that lab will not need apt-mirror.front as
this is only fixing a problem introduced by ceph-qa-chef.


andrewschoen:
I'm not sure we're really gonna need this.  Closing for now.

PR #3, opened 2015-04-28
add puddle role and playbook

ktdreyer:
Add the necessary configuration for installing and running distill and puddle.

Many of the templated variables here are internal to Red Hat and are present in the other "secrets" Ansible repository.

### DIFF
--- a/roles/puddle/tasks/main.yml
+++ b/roles/puddle/tasks/main.yml
@@ -1,0 +1,22 @@
+---
+- fail:
+    msg: "Puddle is only relevant on RHEL"
+  when: ansible_distribution != 'RedHat'
+
+# install puddle
+- include: install.yml
+
+# configure NFS
+- include: nfs.yml
+
+# configure puddle
+- include: configure.yml
+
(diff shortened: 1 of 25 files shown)


roles/puddle/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  when: ansible_distribution != 'RedHat'
+
+# install puddle
+- include: install.yml
+
+# configure NFS
+- include: nfs.yml
+
+# configure puddle
+- include: configure.yml
```
andrewschoen: Tags might be helpful here if we ever want to re-configure puddle without re-installing it.  Thoughts?

andrewschoen:
This is looking great.  I still do think we ought to put the the puddle and distill vars into defaults/main.yml.  I've been trying to make sure every var I add to a group_vars or host_vars file get added to defaults/main.yml as well for the role that uses it.

The puddle and distill vars you put in defaults would obviously have empty values for each key.  Thoughts?

PR #6, opened 2015-04-30
Fixed bug where entitlement setup would fail with --check.

andrewschoen:
The command module is skipped when --check is used which causes the
registered var subscription to not exist the next task fails.


ktdreyer:
:beers:

PR #9, opened 2015-04-30
Adds tags to the ansible-managed role.

andrewschoen:


andrewschoen:
Used https://github.com/ceph/ceph-cm-ansible/pull/14 instead

PR #10, opened 2015-04-30
Adds a new downstream_setup.yml playbook.

andrewschoen:
This playbook will be used to setup testnodes for downstream
testing by downloading arbitrary yum repos. The repos to install should
be passed in with the --extra-vars option to ansible-playbook.

The yum_repos var must be a list of hashes defining the url to download
the repo from and the name to save it as in /etc/yum.repos.d

For example:

      --extra-vars '{"yum_repos":[{"url": "<url>", "name": "name"}]}'

Or, removing repos from /etc/yum.repos.d::

      -- extra-vars "remove_yum_repos=['name']"

### DIFF
--- a/roles/downstream-setup/tasks/yum_repos.yml
+++ b/roles/downstream-setup/tasks/yum_repos.yml
@@ -1,0 +1,7 @@
+---
+- name: Download yum repos to /etc/yum.repos.d
+  get_url:
+    url: "{{ item.url }}"
+    dest: "/etc/yum.repos.d/{{ item.name }}.repo"
+    force: yes
+  with_items: yum_repos
--- a/downstream_setup.yml
+++ b/downstream_setup.yml
@@ -1,0 +1,6 @@
+---
+# A playbook used to setup a node for downstream
+# RHCeph testing.
+- hosts: testnodes
+  roles:
+    - downstream-setup
--- a/roles/downstream-setup/defaults/main.yml
+++ b/roles/downstream-setup/defaults/main.yml
@@ -9,3 +9,7 @@
 #       name: "epel"
 #
 yum_repos: []
+
+# a list of repo names as strings to delete from /etc/yum.repos.d
+# the name should not include the .repo extension
+remove_yum_repos: []
--- a/roles/downstream-setup/defaults/main.yml
+++ b/roles/downstream-setup/defaults/main.yml
@@ -1,0 +1,11 @@
+---
+# yum_repos is a list of hashes that
+# define the url to download the yum repo
+# from and the name to save it as in etc/yum.repos.d
+#
+# For example:
+#   yum_repos:
+#     - url: "http://path/to/epel.repo"
+#       name: "epel"
+#
+yum_repos: []
--- a/roles/downstream-setup/tasks/main.yml
+++ b/roles/downstream-setup/tasks/main.yml
@@ -1,0 +1,4 @@
+---
+- include: yum_repos.yml
+  tags:
+    - yum-repos
--- a/roles/downstream-setup/tasks/main.yml
+++ b/roles/downstream-setup/tasks/main.yml
@@ -1,4 +1,10 @@
 ---
 - include: yum_repos.yml
+  when: yum_repos|length > 0
   tags:
     - yum-repos
+
+- include: remove_yum_repos.yml
+  when: remove_yum_repos|length > 0
+  tags:
+    - delete-yum-repos
--- a/roles/downstream-setup/tasks/remove_yum_repos.yml
+++ b/roles/downstream-setup/tasks/remove_yum_repos.yml
@@ -1,0 +1,6 @@
+---
+- name: Delete yum repos from /etc/yum.repos.d
+  file:
+    path: "/etc/yum.repos.d/{{ item }}.repo"
+    state: absent
+  with_items: remove_yum_repos


roles/downstream-setup/tasks/yum_repos.yml
```diff
@@ -0,0 +1,11 @@
+---
+- fail:
```
andrewschoen: I couldn't decide if failing here or adding a when: yum_repos|length > 0 to the include was a better approach.

If we want this playbook to have multiple snippets somebody building qa-suite yaml could leverage then we'd need some way for that user to chose which snippet to run.  I see two ways. 1) add a when clause to each include to check for the vars needed for the included tasks.  If the vars aren't defined in the yaml, it won't be ran. or 2) support --tags and the user will need to add tags in their yaml specifying which tasks to run.

@zmc @alfredodeza @ktdreyer thoughts?

ktdreyer:
CC'ing @vasukulkarni

roles/downstream-setup/tasks/yum_repos.yml
zmc: I think that gracefully no-oping is far preferable to failing

roles/downstream-setup/tasks/yum_repos.yml
```diff
@@ -0,0 +1,11 @@
...
+    msg: "Please define the repos to install with the var yum_repos."
+  when: yum_repos|length == 0
+
+- name: Download yum repos to /etc/yum.repos.d
+  get_url:
+    url: "{{ item }}"
+    dest: "/etc/yum.repos.d/{{ item }}.repo"
```
andrewschoen: This is bugged now because we changed item to be a URL to a .repo file.  If we want to the flexibility of giving a full URL here we'll also need to provide the name of the repo file to create.  Unless we want to try parsing the url, which is an option but that might get hairy.

andrewschoen:
Here's an example of how I could see this working in ceph-qa-suite yaml.

```
tasks:
- ansible:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    playbook: downstream_setup.yml
    hosts:
      - client.0
    vars:
      yum_repos:
          - url: "http://url/to/epel.repo"
            name: "epel"
- install:
- ansible:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    playbook: downstream_setup.yml
    hosts:
      - client.0
    vars:
      remove_yum_repos:
          - "epel"
```

The first ansible task would download epel.repo, then the install task would run (hypothetically using epel.repo) and finally the ansible task runs again to remove the epel.repo file we created.

The ``Ansible`` task currently doesn't support the ``vars`` argument, but I think it'd be fairly easy to add.

vasukulkarni:
@andrewschoen this looks good to me, I can make the last ansible part of last yaml task that gets run.

ktdreyer:
sounds good

andrewschoen:
@vasukulkarni awesome. If we wanted to get fancy with the ``Ansible`` task another idea I had would be to give the task a list of vars and a playbook to run (from the same repo) during teardown.  This way the first ansible task could cleanup after itself and you wouldn't need to add the additional yaml block at all. @zmc does that idea sound sane?  Maybe something like this.

```
tasks:
- ansible:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    playbook: downstream_setup.yml
    hosts:
      - client.0
    vars:
      yum_repos:
         - url: "http://url/to/epel.repo"
           name: "epel"
    teardown_playbook: downstream_setup.yml
    teardown_vars:
      remove_yum_repos:
         - "epel"
- install:
```

zmc:
@andrewschoen This is looking pretty good!

First, a suggestion for the job yaml: since ``Ansible`` is a ``Task``, it gets overrides for free. You could rewrite your first yaml as:
```
tasks:
- ansible:
- install:
- ansible:
  remove_yum_repos:
     - "epel"

overrides:
- ansible:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    playbook: downstream_setup.yml
    hosts:
      - client.0
    vars:
      yum_repos:
         - url: "http://url/to/epel.repo"
           name: "epel"
```

You could make this slightly slicker by making ``downstream_setup`` instead remove all ``yum_repos`` when ``cleanup`` was set.

I had toyed with the idea of a ``teardown_playbook`` and the more I think about it the more I like it. However, it'd involve a little refactoring. A short-term approach could be to use the first idea (maybe with some overrides), and if ``Ansible`` grows a ``teardown_playbook`` feature it would be easy to migrate.

The one thing I don't really like is having a whole separate ``teardown_vars``; I think if the ansible playbook is well-designed it ought to not even need those.

andrewschoen:
I think we all agree that the current solution proposed here will work for now, right?

@zmc ah! using overrides does make it much nicer here.  Is that second call to the ansible task missing ``vars`` in the override?  Something like:

```
tasks:
- ansible:
- install:
- ansible:
  vars:
    remove_yum_repos:
       - "epel"
```

I hesitate do implement the ``cleanup`` idea right now because I don't want to see these two operations tightly coupled.  I purposely make them separate so that we could inverse this operation if needed.  We might find we need the ability to first remove a repo, add a new one, remove the new one and finally add back the old one. My initial goal for this playbook was to have small snippets of functionality that could be pieced together, allowing for the most flexibility.

Also, wouldn't you still need to pass the yum_repos to delete to the next ``Ansible`` task usage even if you were using some sort of ``cleanup`` var as a trigger?  Would doing something like the following completely replace the ``vars`` in overrides or add to it?  Would the second call there get the ``cleanup`` and ``yum_repos`` vars or just ``cleanup``.

```
tasks:
- ansible:
- install:
- ansible:
  vars:
    cleanup: True

overrides:
- ansible:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    playbook: downstream_setup.yml
    hosts:
      - client.0
    vars:
      yum_repos:
         - url: "http://url/to/epel.repo"
           name: "epel"
```

zmc:
Ah, yeah, vars would be replaced and not extended, good catch. And yes, I did miss the 'vars:' in my example. Everything you're proposing looks fine to me. I'm not attached to using a generic ``cleanup`` var at all; as long as users of the playbook use it correctly, I don't have strong opinions on how vars are passed.

PR #14, opened 2015-05-01
pubkeys and user tags

zmc:
So we can run with e.g. ``--tags="user,pubkeys"``

### DIFF
--- a/roles/ansible-managed/tasks/main.yml
+++ b/roles/ansible-managed/tasks/main.yml
@@ -3,6 +3,8 @@
   group:
     name: sudo
     state: present
+  tags:
+   - user
 
 - name: Create the ansible user.
   user:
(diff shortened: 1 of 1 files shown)


roles/testnode/tasks/user.yml
```diff
@@ -3,6 +3,8 @@
   group:
     name: sudo
     state: present
+  tags:
+   - user
```
andrewschoen: The include in main.yml tags everything in this file as 'user' already.  These tags are unneeded.

roles/ansible-managed/tasks/main.yml
```diff
@@ -24,6 +30,8 @@
     validate: visudo -cf %s
   tags:
     - sudoers
+  tags:
```
andrewschoen: Don't duplicate tags here.  Just add -user underneath of -sudoers

zmc: derp

PR #15, opened 2015-05-01
puddle: add 1.2 configs for ceph-async

ktdreyer:
CEPH-ASYNC advisories have a slightly different config than the config we used for the 1.2.3 release. Here are the changes I made from the 1.2.3 configs:

  1. "errata_release" is "CEPH-ASYNC"
  2. Packages are not always signed (remove the keys settings)
  3. We need to add more repos for repoclosure to succeed (namely, the
     production read-only pulp repos)

### DIFF
--- a/roles/puddle/tasks/configure.yml
+++ b/roles/puddle/tasks/configure.yml
@@ -16,7 +16,9 @@
     mode: 0644
   with_items:
     - ceph-1.2-rhel-6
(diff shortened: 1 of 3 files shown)


ktdreyer:
Crap, I forgot to add these to default.yml :(

ktdreyer:
default vars set in https://github.com/ceph/ceph-cm-ansible/pull/16

PR #17, opened 2015-05-04
Add a teuthology_user group and assign it to the teuthology_user.

andrewschoen:
This fixes a bug when creating a CPAN config directory that belongs to the
group teuthology_user.  If this group wasn't created already that play
fails.

### DIFF
--- a/roles/testnode/tasks/user.yml
+++ b/roles/testnode/tasks/user.yml
@@ -4,6 +4,11 @@
     name: sudo
     state: present
 
+- name: Ensure the teuthology_user group exists.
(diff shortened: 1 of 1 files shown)


zmc:
Should we specify the gid since we're specifying the uid?

andrewschoen:
@zmc I don't think we need too, we didn't for the sudo group anyway.  That said, I'm not opposed to doing that if you think we should.

zmc:
I guess we can hold off on that then; it'll be easy to make that change if we discover that it's necessary.

PR #18, opened 2015-05-05
Cobbler: automatically run ansible after install of new-style profiles

zmc:
For now it only uses ``--tags user,pubkeys``, because I'm not sure if running the whole thing is safe. It might be distro dependant and is definitely something we'd want to test carefully. This should be safe though.

### DIFF
--- a/roles/cobbler/templates/snippets/cephlab_rc_local
+++ b/roles/cobbler/templates/snippets/cephlab_rc_local
@@ -1,0 +1,24 @@
+## {{ ansible_managed }}
+#set lockfile = '/.cephlab_rc_local'
+#set script = '/etc/rc.local'
+
+cat > $script << EOF
+#!/bin/bash
+# First, redirect stderr and stdout to a logfile
+exec 2> /tmp/rc.local.log
+exec 1>&2
+set -ex
+
+# Only run once.
+if [ -e $lockfile ]; then
+    exit 0
+fi
+
+# Wait just a bit for the network to come up
+sleep 15
+# Run the post-install trigger a second time
+wget -O /dev/null "http://$http_server:$http_port/cblr/svc/op/trig/mode/post/system/$system_name"
+touch $lockfile
+EOF
+
+chmod +x $script
--- a/roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
+++ b/roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+## {{ ansible_managed }}
+set -ex
+name=$2
+export USER=root
+export HOME=/root
+ANSIBLE_CM_PATH=/root/ceph-cm-ansible
+
+# Bail if the ssh port isn't open, as will be the case when this is run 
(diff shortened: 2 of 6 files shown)


roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+## {{ ansible_managed }}
+set -ex
+name=$2
+export USER=root
+export HOME=/root
+ANSIBLE_CM_PATH=/root/ceph-cm-ansible
+
+# Bail if the ssh port isn't open.
+nc -vz $name 22
```
andrewschoen: Can we leave some comments about how this is meant to fail silently here the first time it's ran, but then is called again by the rc.local after first reboot?

roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
```diff
@@ -0,0 +1,16 @@
...
+
+mkdir -p /var/log/ansible
+
+pushd $ANSIBLE_CM_PATH
+ANSIBLE_SSH_PIPELINING=1 ansible-playbook testnodes.yml -vv --limit $name* --tags user,pubkeys 2>&1 | tee /var/log/ansible/$name.log
```
andrewschoen: Once we merge https://github.com/ceph/ceph-cm-ansible/pull/1 I think the users.yml playbook would be perfect to use here.  It essentially does what --tags user,pubkeys is doing but will work all for all roles.

ktdreyer:
Just a nit, but it's nice to prefix cobbler-specific commits and PRs with "cobbler:" in the future.

zmc:
Good point @ktdreyer

andrewschoen:
@ktdreyer maybe we should have a standard of 'role: message' for all commits in this repo?

ktdreyer:
definitely

roles/cobbler/templates/snippets/cephlab_rc_local
```diff
@@ -0,0 +1,24 @@
+## {{ ansible_managed }}
+#set lockfile = '/.cephlab_rc_local'
+#set script = '/etc/rc.local'
+
+cat > $script << EOF
+#!/bin/bash
```
andrewschoen: should this be #!/usr/bin/env bash ?

roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
andrewschoen: should this also be #!/usr/bin/env bash ?

ktdreyer: /bin/bash is fine IMHO

zmc: Yeah, /bin/bash works on RHEL 7 and Ubuntu 14.04; if it starts failing on older stuff I'll figure something out.

PR #21, opened 2015-05-05
Cobbler tweaks, and retry entitlements

zmc:


roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
```diff
@@ -13,7 +13,10 @@ nc -vz $name 22
 
 mkdir -p /var/log/ansible
 
-# Tell ansible to create users and populate authorized_keys
 pushd $ANSIBLE_CM_PATH
-ANSIBLE_SSH_PIPELINING=1 ansible-playbook testnodes.yml -vv --limit $name* --tags user,pubkeys 2>&1 | tee /var/log/ansible/$name.log
+export ANSIBLE_SSH_PIPELINING=1
+# Tell ansible to create users and populate authorized_keys
+ansible-playbook testnodes.yml -v --limit $name* --tags user,pubkeys 2>&1 > /var/log/ansible/$name.log
```
andrewschoen: why do we need to do this first if we're gonna run the full playbook anyway?

zmc: When I was seeing the full playbook fail because of the subscription-manager issue, it would fail before really doing anything else, leaving the machine inaccessible. That bugged me so I put user,pubkeys in front.

andrewschoen: That's fair, maybe add --skip-tags user,pubkeys to the next run then?

zmc: Ah, good call!

PR #22, opened 2015-05-06
Use SELinux permissive on RHEL

zmc:


ktdreyer:
For context, this is in support of http://tracker.ceph.com/issues/8731 and http://tracker.ceph.com/issues/9931

PR #24, opened 2015-05-06
Add tools subdir containing vmlist.py script

dmick:

### DIFF
--- a/tools/vmlist.py
+++ b/tools/vmlist.py
@@ -90,7 +90,7 @@ class Cfg(object):
 cfg = Cfg(os.path.expanduser(CONFFILE))
 
 
-def list_vms(host, outputfile):
+def list_vms(host, outputfile=None):
     """
     Connect to host and collect lxc-ls and virsh list --all output
     """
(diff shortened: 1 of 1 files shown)


ktdreyer:
LGTM

dmick:
looking over it, doesn't make a lot of sense to allow one worker to be called with no outputfile but the other one to effectively require it...but, meh.

tools/vmlist.py
```diff
@@ -0,0 +1,198 @@
...
+import sys
+import tempfile
+import textwrap
+
+CACHEFILE = "~/.vmlist.cache"
+CONFFILE = "~/.vmlist.conf"
+
+# mira004.front.sepia.ceph.com is dead
+
+VMMACHINES = textwrap.dedent('''\
```
zmc: How about VM_HOSTS? Also, how about letting it be overridden by setting an env variable of the same name?

zmc:
Curious: why write and re-read outputfiles and not simply use buffers?

tools/vmlist.py
dmick: Name change is fine with me; env vars: well, you can set it in the .conf file too;  I guess we could define an order env -> conf -> default.  You think others of the things should also be in the env, or just the hosts?  (note also the -m switch)

dmick:
outputfiles because multiple processes.  I guess there *are* other methods...Queue and Pipe...but it was not at all clear how to use those, to me, and I wanted to get something working.  I would rather use memory than the filesystem though.

dmick:
Just a poke to get this moving.  Anything else we wanna do to it?

PR #27, opened 2015-05-12
Make rhel 7.0 tasks run on any rhel 7.x version.

andrewschoen:
This shouldn't have been restricted to only 7.0 in the first place. We
noticed this problem while testing rhel 7.1 and noticing that firewalld
was not disabled.


ktdreyer:
LGTM

andrewschoen:
While testing this against a rhel 7.1 node I noticed that nfs-server failed to start because rpcbind wasn't started.  I've pushed a new commit to fix that.

I've tested this against both rhel 7.0, 7.1 and ubuntu trusty (as it also uses nfs.yml) and it's working as expected.

PR #28, opened 2015-05-12
Fix RHEL6

zmc:


roles/cobbler/templates/snippets/cephlab_rhel_disks
```diff
@@ -0,0 +1,15 @@
+## {{ ansible_managed }}
+#set os_version = $getVar('os_version','')
+# Partition clearing information
+clearpart --all --initlabel
+# Use a small root partition on one disk
+part / --fstype="ext4" --ondisk=sda --size=20000
+# Clear the Master Boot Record
+zerombr
+# System bootloader configuration
+#if $os_version == 'rhel7'
```
andrewschoen: Does this condition work for all rhel 7.x versions?

roles/cobbler/templates/snippets/cephlab_packages_rhel
```diff
@@ -1,3 +1,8 @@
 ## {{ ansible_managed }}
+#set os_version = $getVar('os_version','')
```
andrewschoen: I assume this var is set in cobbler somewhere?  Some docs explaining where this comes from might be helpful.

roles/cobbler/templates/snippets/cephlab_rhel_disks
andrewschoen: I think my confusion here is not knowing where/how ``os_version`` is set.

roles/cobbler/templates/snippets/cephlab_packages_rhel
zmc: Yeah, it's from cobbler. Each distro has an "os version" field.

roles/cobbler/templates/snippets/cephlab_rhel_disks
zmc: Yes, e.g. 7.1 still has an os_version of 'rhel7'

roles/cobbler/templates/snippets/cephlab_packages_rhel
andrewschoen: Do you think docs here are a good idea or do we just assume the reader knows enough about cobbler?  It was unclear to me on first read where this came from or what to expect from it.

zmc: It's just a part of the cobbler snippet API; since it's in a cobbler snippet I think it's reasonable to think a user might do something like google "cobbler os_version"

andrewschoen: I think docs would be helpful, but sure.

alfredodeza:
How is it acceptable to expect users to "google" something when adding *one line* to explain what a thing does is not difficult? We should strive for thorough documentation, not less, not 'google it'

roles/cobbler/templates/snippets/cephlab_packages_rhel
zmc: Yeah I'm not sure what to say; we don't usually write docs for APIs we're simply using...

zmc:
@alfredodeza what do you think we should add? Should we be writing docs everywhere we use a cobbler feature? What about other APIs? What about all the little ansible features we use?

I am really confused about how this particular thing is a sticking point.

andrewschoen:
@zmc It's not really a sticking point as the PR is merged.  I do agree with @alfredodeza here though.  I read the code, was confused and did what you expected to find the answer, googled it.  So, I commented to voice that confusion and noted that a few doc lines might help the next reader avoid the same confusion I had.  That's all.

I was unsure if that was a system env var or if it was set in cobbler.  If it's in cobbler, where is it set? Do we set it manually or is it part of the distro import process? Does it return the same values for all minor versions?

alfredodeza:
Yes @zmc we should be writing docs everywhere and as much as possible **especially so** when:

* it is someone else in your team suggesting it
* it is just one line, or two (it doesn't take hours)

It doesn't matter if it is a third party API, or library, or HTTP method. The more the merrier.

zmc:
I'll admit that my response to the original request for docs was too hasty.

If you notice, I used ``$getVar('os_version','')`` twice in this PR. Documenting it both places is clearly not a great thing, especially given the likelihood that we'll end up using stuff like this more often. Additionally, I'm not sure what level of detail such a comment should contain since I don't have a complete understanding of the feature; I discovered it yesterday and it's not well-documented itself (which does bother me).

I can see how it would be beneficial to have documentation that covers stuff like this, but I don't necessarily think inline is the right place. Perhaps a separate document would have been a nice thing to add here.

I'm open to discussing this, but not very excited about doing so in an already-closed PR.

PR #29, opened 2015-05-14
Add cobbler_role profile

zmc:

### DIFF
--- a/roles/cobbler_profile/defaults/main.yml
+++ b/roles/cobbler_profile/defaults/main.yml
@@ -1,9 +1,4 @@
 ---
-# Where to download ISOs
-iso_dir: /var/lib/cobbler/isos
-# Mount point to use for ISOs during import
-iso_mount: /mnt/iso
-
 distros:
   # Distros with empty iso values will be skipped. These dicts will be 
   # updated with same-named items in an 'extra_distros' var, which can be 
--- a/roles/cobbler_profile/defaults/main.yml
+++ b/roles/cobbler_profile/defaults/main.yml
@@ -1,0 +1,22 @@
+---
+# Where to download ISOs
+iso_dir: /var/lib/cobbler/isos
+# Mount point to use for ISOs during import
+iso_mount: /mnt/iso
+
+distros:
+  # Distros with empty iso values will be skipped. These dicts will be 
+  # updated with same-named items in an 'extra_distros' var, which can be 
+  # set in the secrets repo.
+  "RHEL-6.6-Server-x86_64":
+      iso: ""
+  "RHEL-7.0-Server-x86_64":
+      iso: ""
+  "RHEL-7.1-Server-x86_64":
+      iso: ""
+  "CentOS-7-x86_64":
+      iso: http://ftp.linux.ncsu.edu/pub/CentOS/7/isos/x86_64/CentOS-7-x86_64-DVD-1503-01.iso
+      kickstart: cephlab_rhel.ks
+  "CentOS-6.6-x86_64":
+      iso: http://ftp.linux.ncsu.edu/pub/CentOS/6.6/isos/x86_64/CentOS-6.6-x86_64-bin-DVD1.iso
+      kickstart: cephlab_rhel.ks
--- a/roles/cobbler_profile/tasks/import_distro.yml
+++ b/roles/cobbler_profile/tasks/import_distro.yml
@@ -33,7 +33,8 @@
 
 - name: Mount ISO
   command: mount -o loop {{ iso_path }} {{ iso_mount }}
-  when: download|changed
+  when: download|changed or (iso_stat.stat.exists and
+        profile is defined and profile.stdout == '')
   register: mount
 
 - name: Set arch
--- a/roles/cobbler_profile/tasks/import_distro.yml
+++ b/roles/cobbler_profile/tasks/import_distro.yml
@@ -1,0 +1,70 @@
+---
+# This profile will do all the work necessary to create a new distro/profile
+# pair in Cobbler.
+
+# Since this profile will be used several times in the same playbook,
+# mention the distro name each time.
+- name: Distro name
+  debug: var=distro_name
+
+- name: Load extra_distros from secrets
+  set_fact:
+      # The jinja2 templating API allows you to update one dict with
+      # another, but it does so in-place without returning any dict. The
+      # first blob does the updating; the second actually returns the
+      # result.
+      distros: "{{ distros.update(extra_distros|default({})) }}{{ distros}}"
+
+- name: Find distro settings
+  set_fact:
+      distro: "{{ distros[distro_name] }}"
+
+- name: Determine if distro profile exists
+  command: cobbler profile find --name {{ distro_name }}
+  # Skip if the iso field is empty; this allows us to mention distros with 
+  # ISOs that are internal, but leave the URL out.
+  when: distro.iso != ''
+  register: profile
+  ignore_errors: true
+  changed_when: false
+
+- include: download_iso.yml
+  when: distro.iso != ''
+
+- name: Mount ISO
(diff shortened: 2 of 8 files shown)


roles/cobbler_profile/tasks/import_distro.yml
```diff
@@ -0,0 +1,88 @@
...
+  when: distro.iso != ''
+  register: profile
+  ignore_errors: true
+  changed_when: false
+
+- name: Create ISO directory
+  file:
+      path: "{{ iso_dir }}"
+      state: directory
+  when: distro.iso != ''
```
andrewschoen: This same condition is used in quite a few tasks here.  Maybe put all those in another file and only include it when distro.iso != '' ?

roles/cobbler_profile/tasks/import_distro.yml
```diff
@@ -0,0 +1,88 @@
...
+  command: mount -o loop {{ iso_path }} {{ iso_mount }}
+  when: download|changed
+  register: mount
+
+- name: Set arch
+  set_fact:
+      arch: "{{ distro.arch|default('x86_64') }}"
+  when: mount is defined and mount.rc == 0
+
+- name: Import the distro (also creates the profile)
```
andrewschoen: before import you might want to run 'cobbler signature update'.  I had to for it to pick up the new vivid information.  If I didn't the import would fail.

roles/cobbler_profile/tasks/import_distro.yml
```diff
@@ -0,0 +1,88 @@
...
+  command: cobbler import --path={{ iso_mount }} --name={{ distro_name }} --arch={{ arch }}
+  register: import
+  when: mount is defined and mount.rc == 0
+
+- name: Unmount ISO
+  command: umount {{ iso_mount }}
+  when: mount is defined and mount.rc == 0
+
+- name: Check to see if the kickstart needs updating
```
andrewschoen: I believe you can use --kickstart with cobbler import so that the created profile gets the correct kickstart.

"Should you want to force the usage of a specific cobbler kickstart template for all profiles created by an import, you can feed the option --kickstart to import, to bypass the built-in kickstart auto-detection."

From: http://www.cobblerd.org/manuals/2.6.0/3/2/4_-_Import.html

roles/cobbler_profile/tasks/import_distro.yml
zmc: Hrm; it seems not-very-good to run that once per distro. I wonder if there's a way to make sure it's only run once per playbook.

roles/cobbler_profile/tasks/import_distro.yml
zmc: The current behavior allows us to change which kickstart we're using once a profile is created though :)

andrewschoen: Indeed.  Should we do both?  Does it matter?

roles/cobbler_profile/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+# Where to download ISOs
+iso_dir: /var/lib/cobbler/isos
+# Mount point to use for ISOs during import
+iso_mount: /mnt/iso
+
+distros:
```
andrewschoen: seems kinda weird to call this 'distros' when the final result is a profile and the role is called cobbler_profile.

roles/cobbler_profile/tasks/import_distro.yml
andrewschoen: Hmm, it might not hurt anything. But, I suppose that's something the 'cobbler' role could do.

roles/cobbler_profile/tasks/import_distro.yml
```diff
@@ -0,0 +1,88 @@
...
+# pair in Cobbler.
+
+# Since this profile will be used several times in the same playbook,
+# mention the distro name each time.
+- name: Distro name
+  debug: var=distro_name
+
+- name: Load extra_distros from secrets
+  set_fact:
+      distros: "{{ distros.update(extra_distros|default({})) }}{{ distros}}"
```
andrewschoen: I'm confused by the final {{ distros }} there.

roles/cobbler_profile/tasks/import_distro.yml
```diff
@@ -0,0 +1,88 @@
+---
+# This profile will do all the work necessary to create a new distro/profile
+# pair in Cobbler.
+
+# Since this profile will be used several times in the same playbook,
+# mention the distro name each time.
+- name: Distro name
+  debug: var=distro_name
```
andrewschoen: Should we fail if distro_name is not defined?

roles/cobbler_profile/tasks/import_distro.yml
zmc: Eh, I don't see how we'd need both ways; do you?

roles/cobbler_profile/defaults/main.yml
zmc: An actual distribution artifact (ISO) is the input that creates a distro/profile pair. When we reimage, we are acting on system and profile objects; not distros. 

So it's weird, but exactly as weird as the above statement :)

roles/cobbler_profile/tasks/import_distro.yml
zmc: I can add a comment:

The jinja2 templating API allows you to update one dict with another, but it does so in-place without returning any dict. The first blob does the updating; the second actually returns the result.

roles/cobbler_profile/tasks/import_distro.yml
zmc: If ``distro_name`` is not defined, we fail two tasks down in "Find distro settings"

roles/cobbler_profile/tasks/import_distro.yml
andrewschoen: Well, does this actually update the kickstart if we're not importing a new distro?

The when clause here suggests it doesn't.  So, if we want this because it allows us to update the kickstart after the distro has been added we'll need to remove that when clause.

roles/cobbler_profile/tasks/import_distro.yml
andrewschoen: So without the second {{ distros }} it sets distros to an empty string?  If so, that's a bummer.

roles/cobbler_profile/tasks/import_distro.yml
andrewschoen: I guess that's fine, as along as it'll give us a good message why it failed.

I just didn't want this to silently noop or something because of human error.

roles/cobbler_profile/defaults/main.yml
andrewschoen: cobbler *is* weird

roles/cobbler_profile/tasks/import_distro.yml
zmc: It looks like:
```
TASK: [cobbler_profile | Find distro settings] ********************************
fatal: [<redacted>] => One or more undefined variables: 'distro_name' is undefined
```

roles/cobbler_profile/tasks/import_distro.yml
zmc: Yeah, this is maybe something we want in the cobbler role. Maybe even same with the directory-creation i do in ``download_iso.yml`` as well...

andrewschoen: Yeah, 'cobbler signature update' needs to be ran somewhere here.  Cobbler role is good with me.

I'm not as worried about the directory-creation because it will gracefully noop if it's not needed.  The signature update command will be ran on every distro though, which isn't necessary.

zmc: Updated!

PR #30, opened 2015-05-15
Adds the ability to opt-out of enabling epel in the common role.

andrewschoen:
Some of our labs should not be using epel.  This allows us to use
the enable_epel var to opt-out out of that per-lab by using the -secrets
repos.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -12,10 +12,13 @@ epel_repos:
   epel:
     name: "Extra Packages for Enterprise Linux"
     baseurl: "{{ epel_mirror_baseurl }}/{{ ansible_distribution_major_version }}/$basearch"
-    enabled: 1
+    # ternary requires ansible >= 1.9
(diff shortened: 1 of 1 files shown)


ktdreyer:
Awesome, LGTM

ktdreyer:
Zack mentioned today during standup that perhaps this should explicitly disable EPEL if `enable_epel` is false.

andrewschoen:
@ktdreyer @zmc do you mean that we should still download the repo file, but set enabled to 0 or 1 depending on the value of ``enable_epel``.

If so, +1.

ktdreyer:
That's right!

andrewschoen:
Cool, I'll work on that then.

zmc:
Sounds good to me too!

andrewschoen:
@zmc @ktdreyer I've updated this as discussed. The repo file will always be downloaded, but will be enabled or disabled depending on ``enable_epel``.

PR #31, opened 2015-05-15
Install python-devel and python-virtualenv on rhel 7.x

andrewschoen:
python-virtualenv was actually missed in the port, oops.  python-devel was new though.  At some point we should talk again about how we push this kind of stuff out to the task that needs it.  For now, I don't want to hold things up as we figure out the best way to do that.

re: http://tracker.ceph.com/issues/11637

### DIFF
--- a/roles/testnode/vars/redhat_7.yml
+++ b/roles/testnode/vars/redhat_7.yml
@@ -78,5 +78,8 @@ packages:
   - nfs-utils
   # for xfstests
   - ncurses-devel
+  # for s3 tests
+  - python-devel
+  - python-virtualenv
 
 nfs_service: nfs-server


ktdreyer:
s3-tests' `./bootstrap` *ought* to be doing this. I've asked for clarification in that Redmine ticket

andrewschoen:
@ktdreyer It checks, but doesn't install. https://github.com/ceph/s3-tests/blob/master/bootstrap#L17-L28

Should we instead fix that bootstrap?

ktdreyer:
Oh, I didn't notice that. Thanks for pointing that out! https://github.com/ceph/s3-tests/pull/54

PR #32, opened 2015-05-19
puddle: move nginx params on to multiple lines

ktdreyer:
Follow the convention of putting each parameter on multiple lines, instead of cramming everything into one line.

This is a style change only; no functional change.


andrewschoen:
This makes me so happy. Thanks!

PR #34, opened 2015-05-19
puddle: use distill configs from rcm-metadata.git

ktdreyer:
Originally I thought that we could store each of distill's individual configuration files in ceph-cm-ansible.git, but this is becoming untenable because there are too many templated values, and I'm trying to sync these files closely with the copies from Red Hat's rel-eng team.

This commit uses the git module to clone a rcm-metadata.git repository to /etc/distill and alters the "ceph-distill" wrapper script to use that location instead.

Since the individual templated files are no longer needed (the full versions are in rcm-metadata.git), delete them from the tree here.


roles/puddle/tasks/distill/configure.yml
```diff
@@ -1,24 +1,19 @@
...
-    owner: root
-    group: root
-    mode: 0644
-  with_items:
-    - ceph-1.3-rhel-7.conf
-    - comps-ceph-1.3-rhel-7.xml
-    - run-distill.conf
-    - variants-ceph-1.3-rhel-7.xml
+# This repository contains distill's configuration files.
+- name: clone rcm-metadata.git
```
andrewschoen: Do you plan to keep this repo up to date manually after cloning it?  You could probably tag this block, change update to yes and I think it'd handle that for you.

andrewschoen:
This does seem like a much better solution than storing the config files.

roles/puddle/tasks/distill/configure.yml
ktdreyer: Yeah, I figured I would keep the git repo up to date outside of Ansible, because I really want Ansible to be as idempotent as possible. Same thing with using `state=latest` in a yum module... it can end up introducing changes to the system that I didn't explicitly intend to introduce.

andrewschoen: Sounds good to me.

PR #35, opened 2015-05-20
cobbler: one ubuntu preseed for trusty, precise and vivid

andrewschoen:
This adds the new preseed as well as making the necessary changes to cobbler_profile to manage new profiles for precise, trusty and vivid.

Something isn't quite right with the precise and vivid profiles as the don't come up with a network connection.  I suspect it's the kernel options but I'm unsure.  Regardless, I thought it'd be best to put up a PR for this work now and I can follow up with those fixes in a separate PR if I don't figure it out before this is reviewed and merged.


roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
```diff
@@ -79,9 +83,12 @@ d-i user-setup/allow-password-weak boolean false
 d-i user-setup/encrypt-home boolean false
 
 # Individual additional packages to install
+#set os_version = $getVar('os_version','')
```
zmc: Here it looks like you're setting ``os_version`` after you use it in line 20

andrewschoen: Yeah, I should just move this up to the top of the file.

andrewschoen: Just pushed a fix for this.

zmc:
This looks good to me. It's ready to merge from your perspective?

andrewschoen:
@zmc I think so, but let's wait until tomorrow to see if I can figure out that network issue. I also need to submit a PR to the -secrets repo to override kernel_options in typica.

Thanks for the early review though.

andrewschoen:
@zmc Yeah, we should merge this.  After further investigation I don't believe my network issue is related to the kernel_options at all.  We used to be doing this with chef to setup a proper network on ubuntu after being imaged, but we don't have a replacement for this functionality yet: https://github.com/ceph/ceph-qa-chef/blob/master/cookbooks/ceph-qa/recipes/ubuntu.rb#L627-L703

From what I can tell cobbler has a built-in way to do this by setting values on the system being imaged. This is probably the better way to accomplish this anyway as it looks specific to machine types and labs. Maybe the way to fix this is to start managing our systems in cm-ansible as well, which I know we had planned to do eventually anyway.

zmc:
Okay, I'll merge this now and we can handle the network issue later.

Agreed on managing systems in ansible as well.

PR #37, opened 2015-05-21
Testnode: touch /ceph-qa-ready

zmc:
This is something chef does; ``teuthology.task.internal.vm_setup()`` expects it.

### DIFF
--- a/roles/testnode/tasks/main.yml
+++ b/roles/testnode/tasks/main.yml
@@ -64,3 +64,10 @@
 - include: pip.yml
   tags:
     - pip
+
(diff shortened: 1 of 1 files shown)


roles/testnode/tasks/main.yml
```diff
@@ -64,3 +64,8 @@
 - include: pip.yml
   tags:
     - pip
+
```
andrewschoen: Could we add a comment here to indicate why we need this?  I know why, but the next person might not.  The commit log might also be a good place to explain the 'why' of this.

zmc: Done!

PR #39, opened 2015-05-27
Manage CM repos in cobbler role

zmc:
These are used by cobbler after reimaging


ktdreyer:
for clarification, this is something we have already had in place and tested for a while, so we're just making it "official" in ansible here.

ktdreyer:
btw I love `accept_hostkey=true` for an https:// URL, gg Ansible :( I remember discovering that was totally broken for git:// URLs too.

PR #41, opened 2015-05-29
Add kernel_logging tag

zmc:
So that we can easily run just those actions against testnodes

### DIFF
--- a/roles/testnode/tasks/setup-ubuntu.yml
+++ b/roles/testnode/tasks/setup-ubuntu.yml
@@ -30,10 +30,14 @@
 - name: Enable kernel logging to console.
   script: scripts/kernel_logging.sh creates=/kernel-logging-setup
   register: kernel_logging
+  tags:
+    - kernel_logging
 
 - name: Update grub
   command: update-grub
   when: kernel_logging|changed
+  tags:
+    - kernel_logging
 
 - name: Upload ttyS1.conf.
   template:


zmc:
Use like this:
```
ansible-playbook -v testnodes.yml --limit "vpm100*,typica177*" --tags kernel_logging
```

andrewschoen:
This looks reasonable, another option might be to put everything 'kernel_logging' into it's own file and tag that include.  It's only tagging two things, so not a huge deal but if we find ourselves tagging anything else 'kernel_logging' we should probably make it an include.

I'm also wondering if we shouldn't take this chance to rename the script and task here.  I've always thought 'kernel_logging.sh' was a misnomer anyway. I won't block on that if we don't think this is the right PR for that work though.

zmc:
I was confused about the name too but I don't have a better suggestion. As for pulling the tasks out into a new file, I'm not sure how much I care in either direction in this case :)

dmick:
Do we need to tag the dependent action (I.e. is the tag a filter applied before dependency checking)

andrewschoen:
@zmc I don't care to move them to a different file now as it's only two tags but any more than that and I'd advocate it.

andrewschoen:
@dmick it's my understanding that we would need to.  AFAIK, when using tags ansible filters the playbook and *only* runs those tasks that have the tag applied to them.  In this case, if we put those two tasks into a separate include and tagged the include then all tasks in the included file would inherit that tag.

It's worth testing out though.  I know that I had to tag all my include_vars with the tag 'always' so that it would always load vars. Otherwise when saying something like --tags 'entitlements' it wouldn't load the vars needed to properly execute those tasks.

PR #45, opened 2015-06-02
Also correct hostnames on RHEL6

zmc:
Fix them if they are 'localhost' as well.


zmc:
Tested on RHEL 6.6 and 7.1

PR #46, opened 2015-06-02
cobbler: adds cephlab_ubuntu_network snippet

andrewschoen:
This is used to set the network connectivity for ubuntu nodes after
imaging using DHCP.

This is also creating a ``resolvconf`` var that I'm using in the snippet to set resolv.conf and resolv.conf.d/base during imaging.  I need to do that so DHCP will work and give us our IP after first boot.  Eventually we should have the testnodes role use this var as well and delete the hardcoded templates it uses.

### DIFF
--- a/roles/cobbler/defaults/main.yml
+++ b/roles/cobbler/defaults/main.yml
@@ -16,6 +16,7 @@ snippets:
   - cephlab_rc_local
   - cephlab_rhel_disks
   - cephlab_post_install_kernel_options
+  - cephlab_ubuntu_network
 
 scripts:
   - cephlab_preseed_late
@@ -27,3 +28,10 @@ ssh_keys: []
 
 ansible_user: ''
 ansible_user_uid_: ''
+
+# A list of lines to add to resolv.conf and resolv.conf.d/base
+# An example:
+#   resolvconf:
+#      - "nameserver x.x.x.x"
+#      - "search an.example.com"
+resolvconf: []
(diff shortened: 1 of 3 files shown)


zmc:
``/etc/resolv.conf``, in all cases I'm aware of, is auto-generated by something or other (could be NM, could be resolvconf, could be something else). Isn't writing it that file going to just be a band-aid?

andrewschoen:
@zmc I think you're right that something sets it (unsure what) and it does, but it does so incorrectly for precise and vivid. Trusty for whatever reason comes up fine with network without this snippet. Manually creating these files were the only way I could figure out how to resolve this problem.  FWIW, ceph-qa-chef set static IPs and also manually created the resolv.conf files for ubuntu.  Maybe this is just some weirdness with cobbler and ubuntu?

I'm happy to do it a more correct way if anybody knows it.

andrewschoen:
Are we blocking on this until we figure out DHCP in our labs?

roles/cobbler/defaults/main.yml
```diff
@@ -27,3 +28,8 @@ ssh_keys: []
 
 ansible_user: ''
 ansible_user_uid_: ''
+
+# A list of lines to add to resolv.conf and resolv.conf.d/base
+resolvconf:
+  - "nameserver"
+  - "search"
```
zmc: Won't this, as-is, build a broken configuration?

zmc:
No, we were just blocking on me being too busy to remember that I had PRs to look at. I do have one concern, though, which I pointed out in a comment just a second ago.

roles/cobbler/defaults/main.yml
andrewschoen: I'll default this to an empty list and write docs with an example of how to use it.

andrewschoen:
@zmc I just pushed new code for the change to defaults/main.yml that you had suggested.  If ``resolvconf`` is an empty list the template that creates this snippet will not make any changes to resolvconf.

PR #47, opened 2015-06-03
Intelligently enable repo sets

zmc:


roles/common/tasks/main.yml
```diff
@@ -7,6 +7,8 @@
     - empty.yml
   tags:
     - vars
+    # We need these vars for the entitlements tag to work
+    - entitlements
```
andrewschoen: Use the tag 'always'.  This is a magic ansible tag that ensures this will run whenever --tags is used.  I've done this in other places where vars are loaded as well.

If you do that, the 'vars' tag is unneeded.

roles/common/tasks/rhel-entitlements.yml
```diff
@@ -14,26 +14,54 @@
...
+  when: subscription is defined
+
 - name: Register with subscription-manager.
   command: subscription-manager register
            --activationkey={{ subscription_manager_activationkey }}
            --org={{ subscription_manager_org }}
   no_log: true
-  when: subscription is defined and
-        subscription.rc != 0
+  when: rhsm_registered == "False"
```
andrewschoen: Will this error if rhsm_registered is not defined?  I think it would.

roles/common/tasks/rhel-entitlements.yml
```diff
@@ -14,26 +14,54 @@
...
+
+- name: Set replace_repos if newly-subscribed
+  set_fact:
+    replace_repos: True
   when: entitled|changed
 
-- name: enable necessary rhsm repos
+- name: Disable all rhsm repos
+  command: subscription-manager repos --disable '*'
+  when: replace_repos == "True"
```
andrewschoen: Do we need to check to see if replace_repos is defined?  Or maybe we could set a defaults for these sorts of flags instead of always checking to make sure they're defined.

roles/common/tasks/rhel-entitlements.yml
zmc: Sure, but why would it not be defined?

roles/common/tasks/rhel-entitlements.yml
zmc: It will be defined, though.

roles/common/tasks/main.yml
zmc: I'm confused.

I'm adding the entitlements tag so I can entitle machines as a separate step if I need to. I don't necessarily want this to run *always*. The vars tag was added in 5eece205, by you.

roles/common/tasks/rhel-entitlements.yml
andrewschoen: I think I was confused because the set_fact above has a ``when: subscription is defined`` which led me to believe that rhsm_registered might not be created in some circumstances.

roles/common/tasks/rhel-entitlements.yml
```diff
@@ -14,26 +14,54 @@
   changed_when: false
   no_log: true
 
+- name: Set rhsm_registered
+  set_fact:
+    rhsm_registered: "{{ subscription.rc == 0 }}"
+  when: subscription is defined
```
andrewschoen: I believe this is not needed as ``subscription`` looks to always be defined, right?

roles/common/tasks/rhel-entitlements.yml
andrewschoen: Only when ``entitled|changed``. What if we run this against a node that is already registered?  I would guess that ``replace_repos`` would not be defined. Might be wrong though.

roles/common/tasks/main.yml
andrewschoen: Yeah, I quit using the 'vars' tag because I always forgot to include it when using --tags for tasks that required vars to be loaded.  Instead, I've started using ``always`` so I don't have to remember or specially tag include_vars blocks.

Just a suggestion, I find it easier this way.

http://docs.ansible.com/playbooks_tags.html#special-tags

roles/common/tasks/rhel-entitlements.yml
zmc: That's not quite true; it is set if ``repo_list is defined`` right above that.

roles/common/tasks/rhel-entitlements.yml
zmc: True, I think that line can safely be removed.

roles/common/tasks/rhel-entitlements.yml
zmc: Gotcha!

roles/common/tasks/main.yml
zmc: I think if we want to switch from using the vars tag to the always tag, we should do that in a separate PR.

andrewschoen: Alright, do you think this is good to merge then?

zmc: I do

PR #49, opened 2015-06-17
Support for ubuntu vivid

andrewschoen:
There are a few things here I feel need discussion before merge. 

1) We're not going to be maintain a package mirror for vivid and instead will use the distro provided mirror. This was discussed in irc, but more people can weigh in here if they care.

2) There are some packages that we install on other version of ubuntu that aren't available in vivid.  We're already removed mencoder, do we need to find alternate versions for the others commented out in vars/ubuntu_15.yml?

3) Vivid fails to 'enable' nfs-server, but can start it fine.  My first pass at debugging that didn't come up with anything as to why ansible can't 'enable' nfs-server. I'm unsure if this will be an issue or not.

I also have not tested this change against trusty to make sure I didn't break anything so I want to do that before merge as well.  I don't imagine I have though.

### DIFF
--- a/roles/testnode/handlers/main.yml
+++ b/roles/testnode/handlers/main.yml
@@ -15,11 +15,10 @@
     state: started
     enabled: yes
 
-- name: start nfs-server
+- name: restart nfs-server
   service:
     name: "{{ nfs_service }}"
-    state: started 
-    enabled: yes
+    state: restarted
 
 - name: restart cron
   service:
--- a/roles/testnode/tasks/apt_systems.yml
+++ b/roles/testnode/tasks/apt_systems.yml
@@ -1,6 +1,17 @@
 ---
+# Starting with ubuntu 15.04 we no longer maintain our own package mirrors.
+# For anything ubuntu < 15.04 or debian <=7 we still do.
 - name: Setup local repo files.
   include: apt/repos.yml
+  when: ansible_distribution_major_version < 15
+  tags:
+    - repos
+
+- name: Update apt cache.
+  apt:
+    update_cache: yes
+  when: ansible_distribution_major_version >= 15 and
+        ansible_distribution == "Ubuntu"
   tags:
     - repos
 
--- a/roles/cobbler/templates/scripts/cephlab_preseed_late
+++ b/roles/cobbler/templates/scripts/cephlab_preseed_late
@@ -5,7 +5,6 @@ $SNIPPET('cephlab_ubuntu_network')
 # set kernel options as defined by the system, profile or distro
 # in the Kernel Options (Post Install) field which populates the var kernel_options_post
 $SNIPPET('cephlab_post_install_kernel_options')
-$SNIPPET('late_apt_repo_config')
 $SNIPPET('post_run_deb')
 $SNIPPET('download_config_files')
 # custom
--- a/roles/testnode/tasks/nfs.yml
+++ b/roles/testnode/tasks/nfs.yml
@@ -8,4 +8,4 @@
     mode: 0644
   notify:
     - start rpcbind
-    - start nfs-server
+    - restart nfs-server
--- a/roles/testnode/tasks/nfs.yml
+++ b/roles/testnode/tasks/nfs.yml
@@ -9,3 +9,10 @@
   notify:
     - start rpcbind
     - restart nfs-server
+
+- name: Enable nfs-server on rhel 7.x.
+  service:
+    name: "{{ nfs_service }}"
+    enabled: true
+  when: ansible_distribution == "RedHat" and
+        ansible_distribution_major_version == "7"
--- a/roles/testnode/templates/ssh/sshd_config_ubuntu_15
+++ b/roles/testnode/templates/ssh/sshd_config_ubuntu_15
@@ -1,0 +1,91 @@
+# {{ ansible_managed }}
+# Package generated configuration file
+# See the sshd_config(5) manpage for details
+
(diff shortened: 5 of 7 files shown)


roles/testnode/handlers/main.yml
```diff
@@ -19,7 +19,8 @@
   service:
     name: "{{ nfs_service }}"
     state: started 
-    enabled: yes
+# FIXME: vivid fails when trying to enable nfs-server.
+# enabled: yes
```
zmc: Might this be a a bug in ansible or a lack of vivid support on their end?

roles/testnode/tasks/apt_systems.yml
```diff
@@ -1,6 +1,14 @@
 ---
 - name: Setup local repo files.
   include: apt/repos.yml
+  when: ansible_distribution_major_version < 15
```
zmc: Won't debian hit this? Or is that ok?

andrewschoen: Good catch, I'll need to limit this to ubuntu as well.  We only support debian 7 right now, so it would n't be an issue, but better to be explicit anyway though.

roles/testnode/handlers/main.yml
andrewschoen: Let me run it again and see if I can get that error message again.

dmick:
Why are we setting up for vivid if we're not building for it?

andrewschoen:
@dmick I was unclear in my description I think.  We want to test against vivid and we will build ceph packages for it, but we just don't to support / maintain a package mirror for vivid.  i.e. apt-mirror.front

zmc:
Might make sense to update the description

andrewschoen:
I just did, hopefully it's clearer now.

zmc:
Perfect, thanks

roles/testnode/handlers/main.yml
andrewschoen: Here's what I got from ansible after a fresh re-image to vivid.

```
msg: Error when trying to enable nfs-kernel-server: rc=1 Synchronizing state for nfs-kernel-server.service with sysvinit using update-rc.d...
Executing /usr/sbin/update-rc.d nfs-kernel-server defaults
Executing /usr/sbin/update-rc.d nfs-kernel-server enable
Failed to execute operation: No such file or directory
```

Here's the output from ``sudo service nfs-server status`` before reimage.

```
● nfs-server.service - NFS server and services
   Loaded: loaded (/lib/systemd/system/nfs-server.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Wed 2015-06-17 12:30:18 EDT; 6s ago
  Process: 28175 ExecStopPost=/usr/sbin/exportfs -f (code=exited, status=0/SUCCESS)
  Process: 28172 ExecStopPost=/usr/sbin/exportfs -au (code=exited, status=0/SUCCESS)
  Process: 28169 ExecStop=/usr/sbin/rpc.nfsd 0 (code=exited, status=0/SUCCESS)
 Main PID: 23270 (code=exited, status=0/SUCCESS)

Jun 04 12:04:25 magna005 systemd[1]: Starting NFS server and services...
Jun 04 12:04:25 magna005 exportfs[23267]: exportfs: can't open /etc/exports for reading
Jun 04 12:04:25 magna005 systemd[1]: Started NFS server and services.
Jun 04 12:22:01 magna005 systemd[1]: Started NFS server and services.
Jun 17 12:30:18 magna005 systemd[1]: Stopping NFS server and services...
Jun 17 12:30:18 magna005 systemd[1]: Stopped NFS server and services.
```

Here's that same output after re-image:

```
● nfs-server.service - NFS server and services
   Loaded: loaded (/lib/systemd/system/nfs-server.service; enabled; vendor preset: enabled)
   Active: active (exited) since Wed 2015-06-17 13:35:06 EDT; 11min ago
 Main PID: 22132 (code=exited, status=0/SUCCESS)
   CGroup: /system.slice/nfs-server.service

Jun 17 13:35:06 magna005 systemd[1]: Starting NFS server and services...
Jun 17 13:35:06 magna005 exportfs[22129]: exportfs: can't open /etc/exports for reading
Jun 17 13:35:06 magna005 systemd[1]: Started NFS server and services.
```

andrewschoen:
@zmc I've figured out the nfs stuff.  Basically what we want to do is restart nfs for all distros, but then only explicitly enable on rhel 7.x.  This matches the behavior of ceph-qa-chef; at some point I must have decided it was ok to enable it for all distros and that was fine until vivid.

dmick:
but this variance ought to be handled by ansible; 'enabled' is surely generic, even if it has different meanings on different OSes?.  isn't this much of the point of ansible?

zmc:
@dmick yes, but I was wondering if Ansible simply hasn't updated that module to work properly with vivid.

andrewschoen:
@dmick @zmc agreed, ansible should handle this but I also suspect this is just bugged for vivid.  I spent a good amount of time yesterday trying to resolve this and read a bit of the source for the service module, but wasn't able to unravel it.  It might even be something with vivid itself.

That being said, we *never* enabled nfs-kernel-server for ubuntu in the past and it seems to be enabled by default; rhel 7.x was the only distro we explicitly enabled nfs on.

andrewschoen:
I think this is good to go unless anybody else has concerns.

I made an issue in ansible-modules-core about the nfs-kernel-server issue on vivid: https://github.com/ansible/ansible-modules-core/issues/1563

PR #57, opened 2015-06-23
testnode: fixes a bug keeping apt repos from being added

andrewschoen:
I needed to cast ansible_distribution_major_version to an int for the
comparsion to work.


andrewschoen:
Some context, ansible_distribution_major_version is a string, so without casting to an integer the when statement on the include that adds repos would always fail.

This resulted in package install failures for packages that we build.  For example: "No package matching 'libapache2-mod-fastcgi' is available"

PR #63, opened 2015-07-02
Gracefully skip registration

zmc:
If entitlements.yml is not found, don't attempt to register


ktdreyer:
awesome!

PR #64, opened 2015-07-06
Work around BZ#1234967

zmc:
https://bugzilla.redhat.com/show_bug.cgi?id=1234967

### DIFF
--- a/roles/testnode/tasks/yum/packages.yml
+++ b/roles/testnode/tasks/yum/packages.yml
@@ -1,4 +1,9 @@
 ---
+- name: Work around https://bugzilla.redhat.com/show_bug.cgi?id=1234967
+  command: rpm -e ceph-debuginfo
+  tags:
+    - remove-ceph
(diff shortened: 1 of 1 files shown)


ktdreyer:
Mind putting the title of the bug somwhere, so people know whether to click or not? "yum list installed: UnicodeDecodeError"

andrewschoen:
So this works around the bug by ensuring ceph-debuginfo isn't there before we reinstall it?

Also, do you want to remove ceph-debuginfo here then? https://github.com/ceph/ceph-cm-ansible/blob/c9a5c2857a75ef4d95c345b65928706a314a624d/roles/testnode/vars/yum_systems.yml#L9

zmc:
@andrewschoen Yes, it works around the yum bug by using rpm to remove the package.

I didn't remove it from the list because I wasn't sure if it was used elsewhere and it won't cause problems as-is.

PR #67, opened 2015-07-07
Skip ansible_user creation if it is undefined

zmc:

### DIFF
--- a/roles/ansible-managed/tasks/main.yml
+++ b/roles/ansible-managed/tasks/main.yml
@@ -13,6 +13,7 @@
     shell: /bin/bash
     uid: "{{ ansible_user_uid_ }}"
     update_password: on_create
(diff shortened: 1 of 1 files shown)


andrewschoen:
Should we skip it or just provide sane defaults?  Either way, I think ansible_user and ansible_user_uid_ need to be added to a defaults/main.yml for this role.

andrewschoen:
I'm gonna go ahead and merge this and we can follow up with the change to defaults.

zmc:
@andrewschoen Didn't see your question until now, but I think either way would be fine. The inspiration for this PR was teuthology, which doesn't care about ``ansible_user`` at all.

andrewschoen:
@zmc yeah, I'm not sure if we should provide default values or not.  But, I think these two vars should definitely be present in defaults/main.yml is all.  It's no urgent thing.

PR #71, opened 2015-07-09
Use a targets var to optionally define hosts for each playbook

andrewschoen:
This will allow us to pass something like --extra-vars="targets=all"
when running the testnodes playbook to allow it to use any node in the
inventory and not only those in the testnodes group.

I see this as being useful for teuthology so that when the Ansible task
creates an inventory dynamically it doesn't have to worry about trying
to match up the nodes in the inventory with what's defined in the playbook.

### DIFF
--- a/cobbler.yml
+++ b/cobbler.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: cobbler 
+- hosts: "{{ targets | default('cobbler') }}"
   roles:
     - common
     - cobbler 
--- a/puddle.yml
+++ b/puddle.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: puddle
+- hosts: "{{ targets | default('puddle') }}"
   roles:
     - common
     - puddle
--- a/ansible_managed.yml
+++ b/ansible_managed.yml
@@ -1,7 +1,7 @@
 ---
 # a playbook to create the necessary users, groups and
 # sudoer settings needed for ansible to manage a node.
-- hosts: all
+- hosts: "{{ targets | default('all') }}"
   # assuming the nodes we run this on will most likely
   # have an ubuntu user already created.
   vars:
--- a/downstream_setup.yml
+++ b/downstream_setup.yml
@@ -1,6 +1,6 @@
 ---
 # A playbook used to setup a node for downstream
 # RHCeph testing.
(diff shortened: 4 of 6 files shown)


andrewschoen:
corresponding teuthology PR: https://github.com/ceph/teuthology/pull/566

cobbler.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- hosts: cobbler 
+- hosts: "{{ targets | default('cobbler') }}"
```
zmc: I don't think we'd ever need this for cobbler, would we?

puddle.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- hosts: puddle
+- hosts: "{{ targets | default('puddle') }}"
```
zmc: I don't think we'd ever need this for puddle, would we?

ktdreyer: yeah, +1 to skip it for puddle; I think this just makes sense for testnodes

andrewschoen: Agreed.  This is really only an issue for playbooks that teuthology might want to run, but this isn't changing any behavior for cobbler.yml or puddle.yml though so I felt it was harmless.

andrewschoen:
I think we're still trying to decide if we even need this.  Teuthology calls the cephlab.yml playbook from ceph-cm-ansible, but it could very easily recreate that playbook with a python dict and just provide the roles and hosts needed to make this work with a dynamic inventory.

Something like, config['playbook'] = dict(hosts='all', roles=['ansible-managed', 'common', 'testnode'])

dmick:
+1. Much of the semantic of testnodes.yml is its inventory, it seems to me.  "No, use all instead" seems special and layer-violating. If not fully dynamic I'd rather have a different toplevel .yml I think.

andrewschoen:
I was kinda ok with teuthology wiping out 'hosts: testnodes" and using 'hosts: all' because we can be sure that teuthology will only operate against testnodes.  But yes, using testnodes.yml manually with our secrets repos "hosts: testnodes" would need to be used instead to protect the rest of the nodes in our inventory from accidental usage.

andrewschoen:
I don' think we're gonna go this direction. Closing this for now.

PR #72, opened 2015-07-09
implement modify_fstab to skip /etc/fstab modification

dachary:
http://tracker.ceph.com/issues/12250

### DIFF
--- a/roles/testnode/tasks/apt_systems.yml
+++ b/roles/testnode/tasks/apt_systems.yml
@@ -32,6 +32,7 @@
   args:
     creates: /etc/fstab.bak
   register: add_user_xattr
+  when: modify_fstab == true
 
 - name: Enable xattr for this boot.
   command:
--- a/roles/testnode/defaults/main.yml
+++ b/roles/testnode/defaults/main.yml
@@ -22,3 +22,8 @@ xfstests_user: "fsgqa"
 # some distros need to start rpcbind before
 # trying to use nfs while others don't.
 start_rpcbind: true
+
+# set to true if /etc/fstab must be modified to persist things like
+# mount options, which is useful for long lived bare metal machines,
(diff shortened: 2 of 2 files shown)


zmc:
I like the idea of this. However, wouldn't it be way more polite if the logic was inverted here?

In other words, we default to not mangling ``/etc/fstab``, but drop a single boolean into each secrets repo requesting said mangling.

dachary:
@zmc I repushed with modify_fstab instead, as suggested.

roles/testnode/tasks/apt_systems.yml
```diff
@@ -32,6 +32,7 @@
   args:
     creates: /etc/fstab.bak
   register: add_user_xattr
+  when: modify_fstab is defined
```
andrewschoen: modify_fstab will always be defined because it's in defaults/main.yml, so this condition would not make this skip if modify_fstab is false.  All you need here is, 'when: modify_fstab'.

roles/cobbler/defaults/main.yml
```diff
@@ -50,3 +50,8 @@ ansible_user_uid_: ''
 #      - "nameserver x.x.x.x"
 #      - "search an.example.com"
 resolvconf: []
+
+# set to true if /etc/fstab must be modified to persist things like
+# mount options, which is useful for long lived bare metal machines,
+# less useful for virtual machines that are re-imaged before each job
+modify_fstab: true
```
andrewschoen: I just noticed this, sorry.  'modify_fstab' is used in the testnode role, so this should live in roles/testnode/defaults/main.yml

dachary:
http://integration.ceph.dachary.org:8081/ubuntu-2015-07-09_19:53:14-noop-master---basic-openstack/ passes with this change

PR #74, opened 2015-07-14
Fix entitlements

zmc:


roles/common/tasks/rhel-entitlements.yml
```diff
@@ -5,17 +5,19 @@
...
-  include_vars: "{{ entitlements_path }}"
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ entitlements_path }}"
+    - empty.yml
   no_log: true
   tags:
     - vars
-  when: entitlements.stat.exists
+  register: entitlements
```
andrewschoen: Do you use ``entitlements`` anywhere?

zmc: oh, good catch! I had been during development but found a better way to implement the logic I wanted. I'll remove this line.

andrewschoen:
``with_first_found`` is a much nicer approach here, looks good.

PR #75, opened 2015-07-15
Documentation for the testnode role.

andrewschoen:


roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+
+This host is used by teuthology to download ceph packages and will be given higher priority on apt systems::
+
+    gitbuilder_host: gitbuilder.ceph.com
+
+The mirror to download and install ``pip`` from::
+
+    pip_mirror_url: "http://{{ mirror_host }}/pypi/simple"
+
+As hash defining yum repos that would be common across a major version. Each key in the hash will be
```
zmc: *A hash

zmc:
This is fantastic!

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+- Ubuntu Trusty
+- Ubuntu Vivid
+
+**NOTE:** This role was first created as a port of ceph-qa-chef_.
+
+Usage
++++++
+
+The testnode role is primarily used by the ``testnodes.yml`` playbook.  This playbook is run by cobbler during
+bare metal imaging to prepare a node for testing and it also used by Teuthology during test runs to ensure the config
```
dmick: bare-metal

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+is correct before testing.
+
+**NOTE:** ``testnodes.yml`` will only run against nodes in the ``testnodes`` group.
```
dmick: Change to "is limited to run against hosts in the ``testnodes`` group by a hosts: key in the playbook (free opportunity to reinforce how that works)

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+
+Variables
++++++++++
+
+Available variables are listed below, along with default values (see ``defaults/main.yml``). This role also
```
dmick: "(see each role's defaults/main.yml)"

andrewschoen:
Ok, I think I've got all the feedback incorporated here.  I also removed the couple mentions of tags from the ``common`` role.  The ``testnodes.yml`` playbook is the best way to use those tags from the ``common`` role, but I feel those are better documented by that role's README.

roles/testnode/README.rst
```diff
@@ -0,0 +1,206 @@
...
+
+    git_mirror_host: git.ceph.com
+
+This host is used by teuthology to download ceph packages and will be given higher priority on apt systems::
+
+    gitbuilder_host: gitbuilder.ceph.com
+
+The mirror to download and install ``pip`` from::
+
+    pip_mirror_url: "http://{{ mirror_host }}/pypi/simple"
```
dmick: Will this work if mirror_host is undefined (i.e. "don't mirror it")?  Should it?

andrewschoen: Currently this must be defined.  I'd like mirror_host, gitbuilder_host and
pip_mirror_url to all noop eventually if they aren't provided. Then we'd
change all the defaults here to empty strings.

roles/testnode/README.rst
```diff
@@ -0,0 +1,206 @@
...
+
+A hash defining yum repos that would be common across a major version. Each key in the hash will be
+the name of a yum repo created in /etc/yum.repos.d. The key/value pairs as the value for that repo
```
dmick: "name" is a tricky term here.  There's the name of the file, and then the file may contain multiple repos, actually, each with their own [] key, and a separate name= attribute, so there's three names; we should make clear which is which here (and I see that name: key in the actual text).  Documentation is such fun.

roles/testnode/README.rst
```diff
@@ -0,0 +1,206 @@
...
+    # An example: 
+    common_yum_repos:
+      rhel-7-fcgi-ceph:
+        name: "RHEL 7 Local fastcgi Repo"
+        baseurl: http://gitbuilder.ceph.com/mod_fastcgi-rpm-rhel7-x86_64-basic/ref/master/
+        enabled: 1
+        gpgcheck: 0
+        priority: 2
+
+A hash defining version specific yum_repos. Each key in the hash will be
```
dmick: version-specific

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+    # An example:
+    yum_repos:
+      fedora-fcgi-ceph:
+        name: Fedora Local fastcgi Repo
+        baseurl: http://gitbuilder.ceph.com/mod_fastcgi-rpm-fedora20-x86_64-basic/ref/master/
+        enabled: 1
+        gpgcheck: 0
+        priority: 0
+
+A list of packages to install that is specific to a distro version.  These lists are defined in the var files in ``vars/``::
```
dmick: I think maybe a specific example of a particular package that's "specific to a distro version" and one that's "common to a distro or distro version" would be helpful.

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+A list of packages to remove. These lists are defined in the var files in ``vars/``::
+
+    packages_to_remove: []
+
+A list of packages to upgrade. These lists are defined in the vars files in ``vars/``::
+
+    packages_to_upgrade: []
+
+The user that teuthology will use to connect to testnodes. This user will be created by this role and assigned to the appropriate groups.
+Eventhough this variable exists, teuthology is not quite ready to support a configurable user::
```
dmick: Even though

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+
+    teuthology_user: "ubuntu"
+
+This user is created for use in running xfstests from ceph-qa-suite::
+
+    xfstests_user: "fsgqa"
+
+This will control whether or not rpcbind is started before nfs.  Some distros require this, others don't::
```
dmick: Is this based on whether rpcbind is already running?...in which case, can it be automatic?

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+
+    start_rpcbind: true
+
+Set to true if /etc/fstab must be modified to persist things like mount options, which is useful for long lived
```
dmick: long-lived. and I still think there's no reason this couldn't happen on every machine, but...that's another bug

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+
+A list of ntp servers to use::
+
+    ntp_servers:
+      - 0.us.pool.ntp.org
+      - 1.us.pool.ntp.org
+      - 2.us.pool.ntp.org
+      - 3.us.pool.ntp.org
+
+The lab domain to strip from hostname on rhel 7.x nodes. This is only done if ``lab_domain``
```
dmick: why would you want/need to strip a domain from a hostname, and why only for rhel 7.x?

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+user
+    Manages the ``teuthology_user`` and ``xfstests_user``. 
+
+Dependencies
+++++++++++++
+
+This role depends on the following roles:
+
+secrets
+    Provides a var, ``secrets_path``, containing the path to where our secrets are located.
```
dmick: "to where our secrets are located" -> "of the secrets repository, a tree of ansible variable files"

roles/testnode/README.rst
ktdreyer: That or, you know, just move all of this stuff into `defaults.yml` :D

roles/testnode/README.rst
```diff
@@ -0,0 +1,215 @@
...
+    
+sudo
+    Sets ``ansible_sudo: true`` for this role.
```
dmick: ...which causes all the plays "in" (is that the right term) this role to execute with sudo

roles/testnode/README.rst
andrewschoen: I can't remember the exact circumstance for this, but I'm sure we can make this smarter eventually.

roles/testnode/README.rst
andrewschoen: I wish I knew, I suspect it had something to do with how we used to cobbler rhel nodes but don't know for certain.  This is one of the ceph-qa-chef quirks we need to revisit eventually.

PR #76, opened 2015-07-15
Manage the firewall on RHEL/CentOS 6/7

zmc:
On EL7, poke holes for http and https. On EL6, just stop iptables.


ktdreyer:
LGTM

PR #81, opened 2015-07-16
Fix warning in 'when' clause in 'Create the teuthology user.'

dmick:
Warnings appeared each run:

[WARNING]: It is unnecessary to use '{{' in conditionals, leave variables in
loop expressions bare

Apparently 'when' clauses are already Jinja2 expressions, and so don't
need the normal variable quoting.


andrewschoen:
Thanks for this.

PR #82, opened 2015-07-16
Install the Core and Base groups on rpm systems

zmc:
We need to make sure basic things like bzip2 are installed.


dmick:
Do we have a similar issue lurking for debs?

 Why did this just appear now?

andrewschoen:
@dmick my suspicion is that the old pre-baked image we used to use for rhel re-imaging had those there, which is why we didn't see it until we started re-imaging a bunch of nodes the new way which didn't have those (until this patch).

PR #83, opened 2015-07-16
Adds role to install docker

ivotron:
as a submodule


ivotron:
This was done in the context of http://tracker.ceph.com/issues/12379  . As mentioned there, we're going the cgroup route instead of full docker.

PR #85, opened 2015-07-17
testnode: add dbench and bonnie++ for centos7

andrewschoen:
See: http://tracker.ceph.com/issues/12381

### DIFF
--- a/roles/testnode/tasks/yum/packages.yml
+++ b/roles/testnode/tasks/yum/packages.yml
@@ -1,4 +1,10 @@
 ---
+# this is needed for the yum-complete-transation command next
+- name: Ensure yum_utils is present.
+  yum:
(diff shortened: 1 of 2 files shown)


roles/testnode/tasks/yum/packages.yml
```diff
@@ -1,4 +1,10 @@
 ---
+# this is needed for the yum-complete-transation command next
+- name: Ensure yum_utils is present.
+  yum:
+    name: yum-utils
+    state: present
```
zmc: should we use ``latest``?

andrewschoen: I don't have a strong opinion on that, we certainly could.  Seemed to work as 'present' though.

PR #86, opened 2015-07-21
Retry authorized_keys updates

zmc:
http://tracker.ceph.com/issues/12380
We're seeing transient issues connecting to githubusercontent.com. This
commit will cause ansible to retry three times with a five-second delay
between tries. Hopefully that's good enough.

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -43,5 +43,10 @@
     user: "{{ item.name }}"
     key: "{{ item.key }}"
   with_items: managed_users|list + managed_admin_users|list
+  # Register and retry to work around transient githubusercontent.com issues
+  register: ssh_key_update
+  until: ssh_key_update.state == 'present'
+  retries: 3
+  delay: 5
   tags:
     - pubkeys


roles/users/tasks/main.yml
```diff
@@ -43,5 +43,8 @@
     user: "{{ item.name }}"
     key: "{{ item.key }}"
   with_items: managed_users|list + managed_admin_users|list
+  # Register and retry to work around transient githubusercontent.com issues
+  register: ssh_key_update
+  until: ssh_key_update.state == 'present'
```
andrewschoen: I realize the default for ``until`` is 3 retries with a 5 second delay between, but maybe it'd be better to still explicitly define them here?  If nothing else it'd make this block easier to read for somebody who doesn't fully understand how ``until`` works.

andrewschoen:
This looks good to me, just the question on if we should be explicit about the ``retries``and ``delay`` for ``until``.

I also think that in general we should move away from storing links to keys.git in our secrets repos for ssh keys and just store them directly as strings.

roles/users/tasks/main.yml
zmc: I don't mind being explicit about the values; I'll update

zmc:
I agree for the most part, @andrewschoen, but I'd like to see about consolidating the workflow for granting lab access first so that we're not storing them in two different places.

andrewschoen:
@zmc I'd love to see granting lab access be a push button thing.

PR #87, opened 2015-07-21
Cobbler: enable adding more user accounts

zmc:


dmick:
This also fixes an issue with the restart task exposed by testing the user additions.

PR #88, opened 2015-07-22
Cobbler IPMI configuration and utilities

zmc:


roles/cobbler/templates/utils/reimage.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+## {{ ansible_managed }}
+set -ex
+name=$1
+profile=$2
+echo "Reimaging $name with profile $profile"
+cobbler system edit --name=$name netboot off
```
dmick: why set to off first?

roles/cobbler/tasks/main.yml
```diff
@@ -1,4 +1,6 @@
 ---
+- include: ipmi_secrets.yml
```
andrewschoen: I might move the ``always`` tag here instead of having two of them in ipmi_secrets.yml

zmc: True; I had put them there initially in case it needed to be included from a second place. I can move them though.

zmc:
Anything else @andrewschoen ?

andrewschoen:
@zmc nope, just didn't see the update. :)

PR #93, opened 2015-07-24
Cobbler: update CM repos before running ansible

zmc:

### DIFF
--- a/roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
+++ b/roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
@@ -5,6 +5,7 @@ name=$2
 export USER=root
 export HOME=/root
 ANSIBLE_CM_PATH=/root/ceph-cm-ansible
+SECRETS_REPO_NAME={{ secrets_repo.name }}
 
 # Bail if the ssh port isn't open, as will be the case when this is run 
 # while the installer is still running. When this is triggered by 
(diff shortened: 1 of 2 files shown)


zmc:
@andrewschoen what do you think it should default to?

andrewschoen:
@zmc Could make it an empty hash and skip these tasks if it's not provided.  Or a hash defining the keys it needs, but making those empty.

I'm really just trying to minimize vars that seemingly come from nowhere. Docs will help with this as well.

zmc:
@andrewschoen ok. Skipping the tasks wouldn't make sense in this case since we still need the trigger. I can write some bash to be smart about a missing secrets repo, and not try to pull it.

andrewschoen:
@zmc cool, I think we just need to cover the case here where there isn't a secrets repo.  Or, if you do have one docs / an entry in defaults/main.yml to help explain the usage of ``secrets_repo``.

zmc:
@andrewschoen yeah, currently the cobbler roles depend on a secrets repo. Changing that is going to be a lot of work.

roles/cobbler/templates/triggers/install/post/cephlab_ansible.sh
```diff
@@ -13,7 +14,15 @@ nmap -sT -oG - -p 22 $name | grep 22/open
 
 mkdir -p /var/log/ansible
 
+if [ $SECRETS_REPO_NAME != 'UNDEFINED' ]
```
andrewschoen: I think it might be cleaner to use jinja2 templates to handle this condition.  If ``secrets_repo.name == false`` then you could just not include that bit of code at all.  I've done something similar here: https://github.com/ceph/ceph-cm-ansible/blob/master/roles/cobbler/templates/snippets/cephlab_ubuntu_network#L29

I don't want to be pedantic, but I do think this is a good use of jinja2.  Thoughts?

zmc:
I don't disagree here... I'm just not sure if I want to spend ~30-60min re-implementing and testing this just to avoid having a shell evaluate a conditional :-/

andrewschoen:
No big deal, merged.

PR #94, opened 2015-07-27
Adds cgroups packages and net_cls kernel module to ubuntu systems

ivotron:
This is part of the work for http://tracker.ceph.com/issues/12424 which enables the ceph task in ceph/ceph-qa-suite to create and place osd/mon/mds in control groups.

For ubuntu, kernels for precise and up have `CONFIG_CGROUP_*` enabled, so the only thing that it's required is to install `cgroup-lite` which is a service that mounts the cgroups virtual FS. The `cgroup-bin` package will also be used from the ceph task to manipulate (create/modify/delete) control groups.

The `net_cls` subsystem is included as a module (named `cls_cgroup`) and is not enabled loaded by `cgroup-lite`, i.e. it has to be done explicitly. So we add it to `/etc/modules`

### DIFF
--- a/roles/testnode/templates/modules
+++ b/roles/testnode/templates/modules
@@ -8,3 +8,4 @@ loop
 lp
 rtc
 scsi_transport_iscsi
+cls_cgroup
--- a/roles/testnode/vars/ubuntu.yml
+++ b/roles/testnode/vars/ubuntu.yml
@@ -103,6 +103,9 @@ common_packages:
   # nfs
   - nfs-common
   - nfs-kernel-server
+  # cgroups
+  - cgroup-lite
+  - cgroup-bin
 
 packages_to_upgrade:
   - apt


ktdreyer:
The commit message describes what the changes are (the "what" question) but it doesn't answer the question of "why".

Mind adding some more justification to the commit log so we know what these packages are for, or what tests they are for, etc?

ktdreyer:
Looking at your origin branch name I see it is wip-12424, and clicking http://tracker.ceph.com/issues/12424 says "Modify the ceph task to make it possible to put osd/mon/mds in specified control groups."

It would be awesome to include the ticket link and this text in the commit log :)

ivotron:
thanks for the feedback @ktdreyer . I should have marked this as DNM since I was figuring out whether this could potentially cause existing QA tests to fail. I did a lot of reading and asked LXC people to confirm that even though cgroups subsystems are always enabled once cgroup-lite is installed, they should not affect other processes that are not explicitly placed in particular cgroups (besides the addition of some minimal overhead).

Please let me know if there's anything else that is needed in order for this to get merged.

ktdreyer:
Thanks! LGTM

PR #96, opened 2015-08-03
A callback plugin that logs failure message

andrewschoen:
This callback plugin is called on every failure and prints out the
failure information as yaml so that it's easier to read in the
teuthology logs.

If the environment variable ANSIBLE_FAILURE_LOG is present then a log of
all failures in the playbook will be persisted to the file
path given in the ANSIBLE_FAILURE_LOG.

### DIFF
--- a/callback_plugins/failure_log.py
+++ b/callback_plugins/failure_log.py
@@ -1,0 +1,57 @@
+"""
+This callback plugin writes ansible failures to a log as yaml. This way you
+can parse the file later and use the ansible failures for other reporting
+or logging.
+
+A log will not be written unless the environment variable ANSIBLE_FAILURE_LOG
+is present and contains a path to a file to write the log to.
+
+NOTE: This plugin was only tested on ansible version 1.9.2 and does not support
+v2 style plugins
+"""
+import yaml
+import os
+import logging
+
+log = logging.getLogger(__name__)
+# We only want to log if this env var is populated with
+# a file path of where the log should live.
+fail_log = os.environ.get('ANSIBLE_FAILURE_LOG')
+if fail_log:
+    handler = logging.FileHandler(filename=fail_log)
+    log.addHandler(handler)
+
+
+def log_failure(host, result):
+    """
+    If the environment variable ANSIBLE_FAILURE_LOG is present
+    a log of all failures in the playbook will be persisted to
+    the file path given in ANSIBLE_FAILURE_LOG.
+    """
+    if fail_log:
+        failure = {"{0}".format(host): dict()}
+        failure[host] = result
+        log.error(yaml.safe_dump(failure))
+
+
+class CallbackModule(object):
+    """
+    This Ansible callback plugin writes task failures to a yaml file.
+    """
+
+    def runner_on_failed(self, host, result, ignore_errors=False):
+        """
+        A hook that will be called on every task failure.
+        """
+
+        if ignore_errors:
+            return
+
+        log_failure(host, result)
+
+    def runner_on_unreachable(self, host, result):
+        """
+        A hook that will be called on every task that is unreachable.
+        """
+
+        log_failure(host, result)
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.swp
 virtualenv
+*.pyc


andrewschoen:
Produces output like this:

```
TASK: [debug msg="I'll always fail"] ******************************************
failed: [XXX.ceph.redhat.com] => {"failed": true, "failed_when_result": true, "verbose_always": true}
msg: I'll always fail

****Ansible Failure******

XXX.ceph.redhat.com:
  failed: true
  failed_when_result: true
  invocation: {module_args: msg="I'll always fail", module_name: debug}
  msg: I'll always fail
  verbose_always: true
```

ansible.cfg
```diff
@@ -3,3 +3,4 @@ ansible_managed = This file is managed by ansible, don't make changes here - the
 # this works when testing from my laptop, but will need to
 # be changed when it lives in a production environment
 vault_password_file = ~/.vault_pass.txt
+callback_plugins = ~/.ansible/plugins/callback_plugins/:/usr/share/ansible_plugins/callback_plugins:./library/plugins/callback
```
zmc: Why three items here?

library/plugins/callback/failure_log.py
```diff
@@ -0,0 +1,52 @@
...
+
+    If the environment variable ANSIBLE_FAILURE_LOG is present
+    a log of all failures in the playbook will be persisted to
+    the file path given in ANSIBLE_FAILURE_LOG.
+    """
+    failure = {"{0}".format(host): dict()}
+    failure[host] = result
+
+    print "\n****Ansible Failure******\n"
+    print yaml.safe_dump(failure)
```
zmc: So, the above will cause *additional* logging to stdout? I'm not sure if that'll be useful - ansible already logs all this data.

ansible.cfg
andrewschoen: because the other two are the default values and I didn't want to lose those

library/plugins/callback/failure_log.py
andrewschoen: Yeah, ansible logs it, but it's terrible to read.

I liked this output, but I'm willing to remove it if others disagree.

library/plugins/callback/failure_log.py
```diff
@@ -0,0 +1,49 @@
+import yaml
+import os
+
+
+def log_failure(host, result):
+    """
+    Print any failures to stdout nicely formatted as yaml.
```
alfredodeza: I don't see anything here that sends output to stdout

andrewschoen: Whoops, just changed that behavior but forgot to update the docs.

ansible.cfg
zmc: Ah, ok. Maybe a comment would be a good idea here.

library/plugins/callback/failure_log.py
zmc: I'm fine with leaving it in for now - if we all hate it, we can remove it.

ansible.cfg
andrewschoen: Will do.

library/plugins/callback/failure_log.py
andrewschoen: I went ahead and removed it.  Hopefully the nicely formatted message in the exception raised by teuthology will be enough.

andrewschoen:
I've changed this to DNM until I've got this working on the teuthology side as well.

andrewschoen:
This has been tested with teuthology and is working correctly.  My first attempt was not thread safe when multiple nodes were involved, but using the logging module fixes that.

Teuthology PR: https://github.com/ceph/teuthology/pull/589

callback_plugins/failure_log.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def log_failure(host, result):
+    """
+    If the environment variable ANSIBLE_FAILURE_LOG is present
+    a log of all failures in the playbook will be persisted to
+    the file path given in ANSIBLE_FAILURE_LOG.
+    """
+    failure = {"{0}".format(host): dict()}
+    failure[host] = result
```
zmc: Shouldn't these two lines be inside the ``if`` clause below?

andrewschoen: They could be, yeah.  I don't think it would affect the functionality at all though.

dmick: sure, just wasted effort if no logging

andrewschoen: Ok, I've moved those two lines inside the if statement.

zmc: Just less (albeit small) overhead when we're not actually using the feature

PR #98, opened 2015-08-03
DNM: Users: don't change the shell

zmc:
Signed-off-by: Zack Cerza zack@redhat.com

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -23,7 +23,6 @@
   user:
     name: "{{ item.name }}"
     group: sudo
-    shell: /bin/bash
     state: present
   with_items: managed_admin_users
   tags:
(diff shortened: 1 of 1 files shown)


andrewschoen:
You made this addition in a46adef9a0d618b203a32f38dc2ea4a88677dce0, why are we reverting that now?

Also, if we aren't going to set the shell we'll need to update the docs in README.rst to reflect that.

zmc:
Ha, good point @andrewschoen ! I'd forgotten about that commit. I do know that Ubuntu uses ``dash`` as the default shell whereas most other distros use ``bash``.

The inspiration for this is that I use ``zsh`` and don't want that to change every time we run this. But I'll have to find another approach (and update docs). I'll mark this as DNM for now.

andrewschoen:
@zmc The solution I liked for this was adding a shell key to the users list along with the username and ssh key.  This way we can set the shell per user.

PR #100, opened 2015-08-05
Enable EPEL when installing packages on testnodes

zmc:

### DIFF
--- a/roles/testnode/tasks/yum/packages.yml
+++ b/roles/testnode/tasks/yum/packages.yml
@@ -36,6 +36,7 @@
   yum:
     name: "{{ item }}"
     state: present 
+    enablerepo: epel
   with_items: packages
   when: packages|length > 0
 


dmick:
Do you want to do this regardless of the package, or only for some packages?  This seems to run the other risk of unintentional installation of something from epel.

zmc:
Before we disalbed epel on everything, we had the same behavior you're describing. I don't think epel is even allowed to carry packages that are already in RHEL

zmc:
But, this might break CentOS. so... I'm not sure what to do.

ktdreyer:
The issue with EPEL wasn't really about conflicting with RHEL or CentOS itself - it was about conflicting with what we shipped in the "layered product" Red Hat Ceph Storage downstream :(

dmick:
yeah.  We don't want to enable EPEL while installing Ceph, for instance, only some of its dependencies.

I think it would be safer to enable epel only for certain sets of packages, not globally.

zmc:
@dmick that's actually exactly what this PR does.

It also doesn't break CentOS; it fixes it the same way it fixes RHEL. I think this would be good to merge.

dmick:
welll....This looks like it changes the lowlevel worker such that whatever is in the packages variable is allowed to come from epel, which seems ripe for mistakes.  I would think we would want to explicitly say which packages are allowed to come from epel and which are not, such that 'packages' might be split into 'packages' and 'epel_packages' or some such at a higher level.

andrewschoen:
@dmick I agree that's a good approach, do you know of an easy way to know know which list a package would belong in?

Also, I'd love to see teuthology tasks start installing the packages it needs instead of it being done with cm-ansible.  Honestly, I'm not even sure how many packages in that list are really needed.

The immediate need here though is to fix the testnodes role in octo where epel is disabled by default.  We can do that with this PR or by enabling epel again.  I'm hesitant to block on this for too long as octo is pretty broken without it.

andrewschoen:
I've made a list of epel packages for rhel7 in this commit: 3e88d2b3b0c14416ac77d9026514914b9cc4c407

zmc:
Closing in favor of #102

PR #101, opened 2015-08-06
Add the ability to revoke user accounts

zmc:

### DIFF
--- a/roles/users/README.rst
+++ b/roles/users/README.rst
@@ -23,6 +23,10 @@ When adding a user, these steps are performed for each user:
 - Adds the user's public key to ~/.ssh/authorized_keys (tags: pubkeys)
 
 
+This role also supports revoking user access by removing all users in the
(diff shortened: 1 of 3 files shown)


andrewschoen:
This looks good.  I think we need to add the ability to filter the ``revoked_users`` list if the ``users`` var is defined.  We already allow this with ``managed_users`` and ``managed_admin_users``.

zmc:
@andrewschoen When I try the example of filtering users from the README.rst, I get:
```
ERROR: Expecting : delimiter: line 1 column 8 (char 8)
```

andrewschoen:
@zmc oops, typo in the docs.  try:

```
ansible-playbook users.yml -vv --limit="$NODE" --extra-vars='{"users":["user"]}'
```
I'll send a PR to fix that, unless you want to fix it here.

zmc:
@andrewschoen yep, already found and fixed it :)

PR #102, opened 2015-08-06
testnode: put epel packages in their own list for rhel 6 and 7

andrewschoen:
This separates packages we install from epel into their own list for rhel 6 and 7.
We do this so that we can install them with the epel repo explicitly enabled.


zmc:
Looks good to me. @dmick?

dmick:
Sure

PR #103, opened 2015-08-12
roles: add xfsprogs-devel for centos7

jcsp:
Fixes: #12658
Signed-off-by: John Spray <john.spray@redhat.com>

### DIFF
--- a/roles/testnode/vars/centos_7.yml
+++ b/roles/testnode/vars/centos_7.yml
@@ -63,6 +63,7 @@ packages:
   - libacl-devel
   - bc
   - xfsdump
+  - xfsprogs-devel
   ###
   # for blktrace and seekwatcher
   - blktrace


andrewschoen:
Thanks for this!  Eventually we'd like to start pushing the responsibility of installing packages out to the tasks that need them, but we'd like to make that process a bit easier before we start asking for it.

Have you verified that ``xfsprogs-devel`` installs fine on centos7?  The ticket you linked mentions the package name as ``xfsprogs-qa-devel``.

ktdreyer:
`xfsprogs-devel` contains `xfs.h`.
`xfsprogs-qa-devel` contains a lot of other headers.

So I'm thinking the ticket comment was incorrect

jcsp:
like @ktdreyer says, xfsprogs-devel has the header in it.  I've tested that the test is happy once that is installed

PR #105, opened 2015-08-12
Checks were failing due to lack of quotes

djgalloway:
Signed-off-by: dgalloway <dgallowa@REDHAT.COM>

### DIFF
--- a/roles/common/tasks/rhel-entitlements.yml
+++ b/roles/common/tasks/rhel-entitlements.yml
@@ -69,14 +69,14 @@
 
 - name: Disable all rhsm repos
   command: subscription-manager repos --disable '*'
-  when: replace_repos == True
+  when: replace_repos == "True"
   # This produces an absurd amount of useless output
   no_log: True
 
 - name: Enable necessary rhsm repos
   command: subscription-manager repos --enable {{ item }}
   with_items: rhsm_repos
-  when: replace_repos == True and rhsm_repos|length > 0
+  when: replace_repos == "True" and rhsm_repos|length > 0
 
 - name: Remove old apt-mirror repository definition.
   file:


roles/common/tasks/rhel-entitlements.yml
```diff
@@ -69,14 +69,14 @@
 
 - name: Disable all rhsm repos
   command: subscription-manager repos --disable '*'
-  when: replace_repos == True
+  when: replace_repos == "True"
```
andrewschoen: I think I'd prefer is this was just ``when: replace_repos``.

roles/common/tasks/rhel-entitlements.yml
```diff
@@ -69,14 +69,14 @@
...
   # This produces an absurd amount of useless output
   no_log: True
 
 - name: Enable necessary rhsm repos
   command: subscription-manager repos --enable {{ item }}
   with_items: rhsm_repos
-  when: replace_repos == True and rhsm_repos|length > 0
+  when: replace_repos == "True" and rhsm_repos|length > 0
```
andrewschoen: Same comment as above.

dmick: Is string truth really right?   Is there no Boolean?  Also, is "a" true?

andrewschoen: We explicitly set ``replace_repos`` to either True or False.  From my understand (true, True, False and false) will all be made a boolean by jinja2.  If we wanted we could do an explicit cast to a bool with: ``when: "{{ replace_repos|bool }}"`` I believe.

See: http://docs.ansible.com/ansible/playbooks_conditionals.html#the-when-statement

PR #106, opened 2015-08-12
Wip rhel entitlements

djgalloway:
Bug was found by re-imaging RHEL7.1 system while customer portal was down, and thus, subscription-manager failed.
Ran testnodes playbook again after outage resolved and rhsm_repos were not disabled.
This resulted in yum-plugin-priorities installation failure due to missing item=rhel-7-server-optional-rpms repo
Remander of playbook failed at that point

### DIFF
--- a/roles/common/tasks/rhel-entitlements.yml
+++ b/roles/common/tasks/rhel-entitlements.yml
@@ -52,7 +52,7 @@
     repo_list: "{{ repo_list_cmd.stdout.split('\n') }}"
   when: repo_list_cmd is defined and not repo_list_cmd|skipped
 
-- name: Set replace_repos if rhsm_repos changed
+- name: Set replace_repos if rhsm_repos differs from repo_list
   set_fact:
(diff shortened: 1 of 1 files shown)


djgalloway:
`Set replace_repos if entitlement is skipped` would set `replace_repos` as `False` if the system was subscribed but its repos needed fixing.

In this commit, either the `Set replace_repos if rhsm_repos differs from repo_list` or `Set replace_repos if newly-subscribed` task will always run and return `True` or `False` for `replace_repos`.

I verified this by running the task against a RHEL7.1 system using the following scenarios:
* unsubscribed (`Set replace_repos if newly-subscribed` sets  `replace_repos: True`)
* subscribed with incorrect repos (`Set replace_repos if rhsm_repos differs from repo_list` sets `replace_repos: True`)
* subscribed with correct repos (`Set replace_repos if rhsm_repos differs from repo_list` sets `replace_repos: False`)

andrewschoen:
@djgalloway Ok, sounds good.  Squash up the commits and this is good to go.

PR #109, opened 2015-08-14
python-rados does not exist in firefly

dachary:
It must be removed if present otherwise a firefly installation will
fail with:

Package python-ceph is obsoleted by python-rados,
  but obsoleting package does not provide for requirements

or similar errors.

Signed-off-by: Loic Dachary <loic@dachary.org>

### DIFF
--- a/roles/testnode/vars/apt_systems.yml
+++ b/roles/testnode/vars/apt_systems.yml
@@ -9,6 +9,7 @@ ceph_packages:
   - libcephfs1
   - radosgw
   - python-ceph
+  - python-rados
   - librbd1
   - librados2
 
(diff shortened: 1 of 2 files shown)


dachary:
@smithfarm : if that is a way to ensure removal of python-rados prior to using the machine, it would solve the http://pulpito.ceph.com/smithfarm-2015-08-10_12:35:46-powercycle-firefly-backports-testing-basic-multi/ bug. I'm not entirely sure it's correct though. Still very new to ansible ;-)

smithfarm:
Only post-firefly versions of ceph have a `python-rados` package. The RPM spec file of these later versions specifies for `python-rados` that:

    Obsoletes:      python-ceph < %{epoch}:%{version}-%{release}

. . . implying that if `python-rados` is installed, it will obsolete all older versions of `python-ceph`, effectively preventing the firefly (0.80.10) version of `python-ceph` from being installed.

andrewschoen:
Wow, I completely forgot to document ``ceph_packages``.  I'll fix that.

This looks good, thanks!

PR #110, opened 2015-08-14
tesnode: add `ceph_packages` to the README

andrewschoen:
I just forgot to document this the last time around.


dmick:
Why isn't this named ceph_packages_to_remove?

dmick:
Also, is it safe with all pkgmgrs to specify packages that aren't installed and/or don't exist?  If so that's another good clause to add to docs

andrewschoen:
@dmick  I've pushed a commit that renames ``ceph_packages`` to ``ceph_packages_to_remove``.

It is safe.  I added a bogus package name to the list and ran the playbook against trusty and centos 7.  I'll add that to the docs.

andrewschoen:
I've added a sentence to the README mentioning that both ``yum`` and ``apt-get`` handle the case where a package isn't installed or doesn't exist without failing loudly.

PR #111, opened 2015-08-14
testnode: make teuthology_user passwordless

andrewschoen:
Fixes: http://tracker.ceph.com/issues/12670

### DIFF
--- a/roles/testnode/tasks/user.yml
+++ b/roles/testnode/tasks/user.yml
@@ -22,6 +22,13 @@
   # If we're currently running as teuthology_user, we won't be able to modify
   # the account
   when: teuthology_user != ansible_ssh_user
+  register: teuthology_user_existence
+
+# If the teuthology_user was just created, delete its password
+- name: Delete the teuthology users password.
+  command: "passwd -d {{ teuthology_user }}"
+  when: teuthology_user_existence is defined and
+        teuthology_user_existence|changed
 
 - name: Add a user for xfstests to test user quotas.
   user:


andrewschoen:
@dmick anymore questions on this?

dmick:
I had asked some more in chat but I don't recall seeing an answer...  My confusion was basically what the difference was between "the variable is defined" and "the variable passed the 'changed' filter", because they seemed redundant: either it's defined and changed, or not defined and therefore changed is irrelevant, so why isn't "defined" enough.

andrewschoen:
@dmick I attempted to answer here: https://github.com/ceph/ceph-cm-ansible/pull/111#commitcomment-12716482

"defined" is not enough because we only want to wipe the password if it's the first time we created the teuthology_user.  Does that make sense?  We can't just used the 'changed' filter either because 'Create the teuthology user' could be skipped and then we'd be checking that status of an undefined var, which makes the playbook explode.

dmick:
"only want to wipe....the first time".  I guess that's my point...won't there ever only *be* a "first time"?  I.e., is it possible for the play to run to the point of defining the variable, but *not* create the user?

andrewschoen:
@dmick Yes, anytime a play is run it will register the result of the play in the variable you specify, even if play did absolutely nothing.  In that case ``teuthology_user_created|changed == False``.

andrewschoen:
@dmick sure thing, I'll make that change

andrewschoen:
@dmick I've just updated this with your suggestion, I also added a comment I thought might help.

PR #116, opened 2015-08-20
downstream-setup: separate cleanup and setup tasks into their own files

andrewschoen:
This is a refactor that makes this role a bit easier to understand and
use. I've simplified main.yml to only include setup.yml and cleanup.yml.
This separation of cleanup and setup related tasks make it a bit easier
to understand.

Also fixes this bug: http://tracker.ceph.com/issues/12693

### DIFF
--- a/roles/downstream-setup/tasks/main.yml
+++ b/roles/downstream-setup/tasks/main.yml
@@ -1,36 +1,14 @@
 ---
-- include: yum_repos.yml
-  when: yum_repos|length > 0 and
-        not cleanup
-  tags:
-    - yum-repos
+# These are tasks which perform actions corresponding to the names of
+# the variables they use.  For example, `disable_yum_repos` would actually
+# disable all repos defined in that list.
+- include: setup.yml
+  when: not cleanup
 
-- name: Set remove_yum_repos on cleanup
-  set_fact:
-    remove_yum_repos: "[{% for repo in yum_repos %}'{{ repo.name}}',{% endfor %}]"
-  when: yum_repos|length > 0 and
-        cleanup and
-        remove_yum_repos|length == 0
-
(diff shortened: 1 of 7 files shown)


roles/downstream-setup/tasks/main.yml
```diff
@@ -1,36 +1,14 @@
...
-  when: disable_yum_repos|length > 0 and
-        cleanup and
-        enable_yum_repos|length == 0
-
-- include: enable_yum_repos.yml
-  when: enable_yum_repos|length > 0
-  tags:
-    - enable-yum-repos
+# These are tasks where the variables used perform the action
+# defined by their name.  For example, `disable_yum_repos` would actually
```
zmc: This feels like strange wording to me. Maybe something like "These are tasks which perform actions corresponding to the names of the variables they use" ?

The other paragraph could say "tasks which reverse the actions corresponding to..."

roles/downstream-setup/tasks/main.yml
```diff
@@ -1,36 +1,14 @@
...
+# disable all repos defined in that list.
+- include: setup.yml
+  when: not cleanup
+
+# These are tasks where the variables used perform the inverse action
+# defined by their name.  For example, `disable_yum_repos` would actually
+# enable all repos defined in that list. The primary use for this is through
+# teuthology, so that you can tell a teuthology run to disable a set of repos
+# for the test run but then renable them during the teuthology cleanup process.
```
zmc: re-enable

zmc:
Looks great; just some docs pickiness

andrewschoen:
Ok, cool.  I've got those docs updates made.

zmc:
Thanks!

PR #117, opened 2015-08-20
Enable mirrorlist for epel repo

djgalloway:
This enables mirrorlist functionality with the epel repo.

This should prevent package installation from breaking when yum runs into a connection timeout with the baseurl.


djgalloway:
Resolves http://tracker.ceph.com/issues/12739

PR #120, opened 2015-08-26
testnodes: Check for and remove custom apt repos

zmc:
http://tracker.ceph.com/issues/12794

### DIFF
--- a/roles/testnode/tasks/apt/repos.yml
+++ b/roles/testnode/tasks/apt/repos.yml
@@ -1,4 +1,17 @@
 ---
+# Check for and remove custom repos.
+# http://tracker.ceph.com/issues/12794
+- name: Check for custom repos
+  shell: "ls -1 /etc/apt/sources.list.d/"
+  register: custom_repos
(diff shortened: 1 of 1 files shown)


roles/testnode/tasks/apt/repos.yml
```diff
@@ -1,4 +1,17 @@
 ---
+# Check for and remove custom repos.
+# http://tracker.ceph.com/issues/12794
+- name: Check for custom repos
+  shell: "ls -1 /etc/apt/sources.list.d/"
+  register: custom_repos
+  changed_when: false
+
+- name: Remove custom repos
```
andrewschoen: Does this need a ``when``?  What happens if there are no repos in /etc/apt/sources.list.d?

zmc: It doesn't; ``with_items`` will iterate over the empty list and do nothing

andrewschoen: Ok, perfect.

PR #124, opened 2015-09-01
testnode: retry when updating the apt cache to handle http errors

andrewschoen:
When updating the apt cache we have to make an http call out to
gitbuilder, if gitbuilder fails for whatever reason we should retry
before failing the playbook.

Also, this removes a 'changed_when: false' on 'Add local apt repos'.
We check it's 'changed' status to know if we need update the cache or
not. We only added that 'changed_when' to avoid seeing this task being
changed each time (because it will always change), but we actually need
that value set correctly because we use it in the 'when' clause of the
updating apt cache stanza.


andrewschoen:
See: http://tracker.ceph.com/issues/12900

PR #131, opened 2015-09-10
Pulpito role

zmc:


andrewschoen:
Overall I'd say this looks pretty good.  I left a few comments and I think the commit messages could be a bit more descriptive.  Good stuff, though.

zmc:
@andrewschoen should be good to go!

PR #132, opened 2015-09-15
Updates to nagios disk monitoring script

djgalloway:
Adds support for magnas / non-raid systems
Adds support for nagios multiline output

Signed-off-by: dgalloway <dgallowa@REDHAT.COM>

### DIFF
--- a/roles/testnode/files/libexec/smart.pl
+++ b/roles/testnode/files/libexec/smart.pl
@@ -9,6 +9,7 @@ my $crit;
 my $out;
 
 my @out;
+my @failedout;
 my $drives;
 my $pci;
 my $type;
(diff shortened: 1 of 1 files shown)


djgalloway:
See mira026 in nagios for sample Service output

dmick:
this would be much better if we split it into logical chunks/commits by theme/fix/enhancement.  We can work together to do that if you like.

zmc:
@dmick anything holding this back? Not sure if we want to block this on a rewrite, but if you feel strongly... ;)

dmick:
I hadn't yet understood the answer to "why remove $scsi".  It makes sense now I think.  And bleh about the array counts; it just increases my desire to rewrite this.

PR #137, opened 2015-09-21
rm ceph-extras repository config

ktdreyer:
This repository contains packages that we don't have time to continue maintaining. In particular, the CentOS 6 QEMU packages were vulnerable to VENOM (CVE-2015-3456) and no users should rely on them. The other packages in "ceph-extras" will continue to incur security issues over time.

With the recent migration from "ceph.com" -> "download.ceph.com", we've chosen not to copy these "ceph-extras" files over to the new locations.

### DIFF
--- a/roles/testnode/vars/centos_6.yml
+++ b/roles/testnode/vars/centos_6.yml
@@ -2,12 +2,6 @@
 # vars specific to centos 6.x
 
 yum_repos:
-  centos6-qemu-ceph:
-    name: Cent OS 6 Local Qemu Repo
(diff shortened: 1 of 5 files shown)


zmc:
Thanks @ktdreyer!

ktdreyer:
To anyone who may still experience `apt-get update` errors on Ubuntu Precise in our labs, please delete the following file:
```
sudo rm /etc/apt/sources.list.d/ceph_com_packages_ceph_extras_debian.list
```

PR #140, opened 2015-09-23
testnode: add fcgi to the list of EPEL packages

ktdreyer:
For Ceph upstream, we rely on fcgi from EPEL.

http://tracker.ceph.com/issues/13203 Refs: #13203


liewegas:
lgtm, but i'm not the expert

PR #142, opened 2015-09-24
Use local epel mirrorlist

djgalloway:
Related to:
http://tracker.ceph.com/issues/12739
http://tracker.ceph.com/issues/12778

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -9,12 +9,11 @@ rhsm_repos: []
 kerberos_realm: EXAMPLE.COM
 
 epel_mirror_baseurl: "http://dl.fedoraproject.org/pub/epel"
-epel_mirrorlist: "http://mirrors.fedoraproject.org"
 epel_repos:
   epel:
     name: "Extra Packages for Enterprise Linux"
     baseurl: "{{ epel_mirror_baseurl }}/{{ ansible_distribution_major_version }}/$basearch"
-    mirrorlist: "{{ epel_mirrorlist }}/metalink?repo=epel-{{ ansible_distribution_major_version }}&arch=$basearch"
+    mirrorlist: file:///etc/yum.repos.d/epel-mirrorlist
     failovermethod: priority
     # ternary requires ansible >= 1.9
     enabled: "{{ enable_epel | ternary(1, 0) }}"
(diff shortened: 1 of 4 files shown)


ktdreyer:
I didn't know you could put newlines in like that?

djgalloway:
Tested it on plana11.  See [[repository] OPTIONS](http://linux.die.net/man/5/yum.conf) section.

The other way to maintain this would be to create separate local files containing mirror URLs

roles/common/defaults/main.yml
```diff
@@ -9,20 +9,17 @@ rhsm_repos: []
 kerberos_realm: EXAMPLE.COM
 
 epel_mirror_baseurl: "http://dl.fedoraproject.org/pub/epel"
-epel_mirrorlist: "http://mirrors.fedoraproject.org"
 epel_repos:
   epel:
     name: "Extra Packages for Enterprise Linux"
-    baseurl: "{{ epel_mirror_baseurl }}/{{ ansible_distribution_major_version }}/$basearch"
-    mirrorlist: "{{ epel_mirrorlist }}/metalink?repo=epel-{{ ansible_distribution_major_version }}&arch=$basearch"
+    baseurl: "{{ epel_mirror_baseurl }}/{{ ansible_distribution_major_version }}/$basearch\nhttp://mirrors.cat.pdx.edu/epel/{{ ansible_distribution_major_version }}/$basearch\nhttp://mirror.pnl.gov/epel/{{ ansible_distribution_major_version }}/$basearch\nhttp://fedora-epel.mirror.lstn.net/{{ ansible_distribution_major_version }}/$basearch"
```
andrewschoen: Could we instead have ansible manage a file, put that file on the remote and then point to it here in baseurl?

All this put into one string is difficult to parse, IMHO.

ktdreyer: Agreed, a separate file sounds a lot saner

zmc: A template might be useful here:
http://docs.ansible.com/ansible/template_module.html

djgalloway:
Suggestions on a cleaner way to do this are welcome

roles/common/tasks/epel.yml
```diff
@@ -16,6 +16,12 @@
   register: epel_repo
   with_dict: epel_repos
 
+- name: Configure local epel mirrorlists
+  template: src={{ item }} dest=/etc/yum.repos.d/{{ item }} owner=root group=root mode=0644
```
andrewschoen: nitpick: I prefer this kind of formatting:

```
template:
  src: "{{ item }}"
  dest: "/etc/yum.repos.d/{{ item }}"
...
```

I think it's easier to read and makes for better diffs if/when we ever change this code.

andrewschoen:
All I had was a style nitpick, everything else here looks good to me.

ktdreyer:
thanks @djgalloway !

PR #143, opened 2015-09-25
Set hostname in cobbler system profiles

djgalloway:
Ran across this and remembered someone saying, at some point, "seems like something we should set in ansible."

### DIFF
--- a/roles/cobbler_systems/tasks/populate_systems.yml
+++ b/roles/cobbler_systems/tasks/populate_systems.yml
@@ -14,7 +14,7 @@
       "{{ groups.cobbler_managed | difference(cobbler_systems_current) }}"
 
 - name: Add missing systems to cobbler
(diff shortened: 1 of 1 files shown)


zmc:
This looks good but I want to test it first.

The commit message: "Should clobber bug we ran into" doesn't parse - could you rephrase it please?

zmc:
Turns out this does the following:

```
changed: [plana32.front.sepia.ceph.com] => (item=burnupi04.front.sepia.ceph.com) => {"changed": true, "cmd": ["cobbler", "system", "edit", "--name=burnupi04", "--mac=e8:9a:8f:91:bb:e4", "--ip-address=", "--interface=eth0", "--hostname=plana32.front.sepia.ceph.com", "--kopts=", "--power-type=ipmilan", "--power-address=burnupi04.ipmi.sepia.ceph.com", "--power-user=inktank", "--power-pass=ApGNXcA7"], "delta": "0:00:00.510144", "end": "2015-10-01 15:44:52.375687", "item": "burnupi04.front.sepia.ceph.com", "rc": 0, "start": "2015-10-01 15:44:51.865543", "stderr": "", "stdout": "", "warnings": []}
```

Really doesn't seem like what we want.

djgalloway:
@zmc Hm..  Would it be better to change to `{{ item }}` or maybe `{{ item.split('.')[0] }}.{{ lab_domain }}`?

zmc:
*probably* the latter would be the right choice

djgalloway:
@zmc Guess I never commented that I updated this.  Should be OK to merge.

zmc:
@djgalloway Looks good. I'm testing it right now - downloading isos to new-sepia-cobbler is taking ages.

PR #144, opened 2015-09-25
Remove custom qemu packages

djgalloway:
This commit removes installation requirement of custom Ceph qemu packages that are no longer available for RHEL/CentOS6 and shouldn't be used on RHEL/CentOS7.

Repos containing these packages were removed in https://github.com/ceph/ceph-cm-ansible/commit/0d3dcfe3b3d85901016d6bb767ef17b5c791817e


ktdreyer:
LGTM! :beer:

PR #145, opened 2015-09-30
testnode: set_hostname should be a global task

djgalloway:
Newly provisioned CentOS testnodes were being left with hostname 'localhost'

This task should apply to all distros; not just RHEL.


zmc:
We should probably update the comment in ``roles/testnode/defaults/main.yml`` to reflect this, right?

djgalloway:
@andrewschoen Do you remember what the specific quirk was mentioned in https://github.com/ceph/ceph-cm-ansible/commit/6701c31497c85357a563640fc4b3a2d2541b4772?

Was it maybe a result of [issue 12672](http://tracker.ceph.com/issues/12672)?

Either way, I still think lab_domain is a useful variable.  I think it's just a matter of redefining its use now.

PR #158, opened 2015-10-26
testnodes: Add teuthology_user to 'disk' group

zmc:
We were already doing this for apt systems; do it for yum systems as
well


zmc:
For reference: https://github.com/ceph/ceph-cm-ansible/blob/master/roles/testnode/tasks/apt_systems.yml#L58-L64

idryomov:
This needed for krbd, +1.

PR #160, opened 2015-10-27
vmlist.py: query multiple openstack providers, 'regions' in provider, add .conf template

dmick:


dmick:
added a fourth commit to do "Add cloud-.." better

dmick:
OK, this is in better form, zack, if you wanna have a peek

PR #161, opened 2015-10-29
cobbler: Updating dban profile

djgalloway:
Since we added dban to the labs to automatically wipe disks before decommissioning a host, I thought it'd be fitting to change the dban cobbler profile accordingly.


djgalloway:
Taking care of this in secrets repos instead

PR #163, opened 2015-11-03
testnodes: Set hostnames properly

zmc:
Instead of using convoluted logic to determine whether or not to set the
hostname, just do it. The hostname module is nice and idempotent.

This was inspired by http://tracker.ceph.com/issues/13679


zmc:
http://tracker.ceph.com/issues/13679#note-2

PR #166, opened 2015-11-12
testnode: do not install fastcgi on RHEL

ktdreyer:
Red Hat QE occasionally tests full in-place OS upgrades from RHEL 6 to
RHEL 7. When mod_fastcgi is pre-installed on a RHEL 6 node, and the node
is upgraded to RHEL 7, then Apache will refuse to start unless the QE
team member also upgrades the mod_fastcgi package explicitly. (see
https://bugzilla.redhat.com/1230679 for an example)

Since RGW's Civetweb is becoming mainstream, we don't really need to
focus on mod_fastcgi for Ceph any more, and having it pre-installed on
our RHEL nodes is more of a hindrance than a help.

I'm leaving the CentOS configs alone, for now, so we don't interfere
with any upstream testing with mod_fastcgi.

### DIFF
--- a/roles/testnode/vars/redhat_6.yml
+++ b/roles/testnode/vars/redhat_6.yml
@@ -8,12 +8,6 @@ common_yum_repos:
     enabled: 1
     gpgcheck: 0
     priority: 2
-  centos6-fcgi-ceph:
-    name: "Cent OS 6 Local fastcgi Repo"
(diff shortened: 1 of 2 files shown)


zmc:
Looks good to me

vasukulkarni:
:+1:

PR #167, opened 2015-11-18
rhel package changes required for ceph

vasukulkarni:
Rhel packages that needs to be added or removed to satisfy dependency and work across various ceph releases without conflict, Still in testing.

### DIFF
--- a/roles/testnode/tasks/yum/packages.yml
+++ b/roles/testnode/tasks/yum/packages.yml
@@ -32,6 +32,15 @@
   tags:
     - remove-ceph
 
+- name: Ensure ceph dependency packages are not present.
+  yum:
+    name: "{{ item }}"
+    state: absent
+  with_items: ceph_dependency_packages_to_remove
+  tags:
+    - remove-ceph-dependency
+
+
 - name: Install packages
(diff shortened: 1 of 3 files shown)


vasukulkarni:
Dependent on https://github.com/ceph/teuthology/pull/709

zmc:
Will this cause problems on CentOS/Fedora? Should we make it only affect RHEL?

zmc:
Also please tweak the commit message of the first commit so it doesn't overflow

vasukulkarni:
@zmc: this should not create issues for centOS and fedora, it will help by removing those dependency present and if present can create conflict. I think we should do this for apt as well.

vasukulkarni:
@zmc fixed commit messages

andrewschoen:
@vasukulkarni would you mind rebasing this branch?

vasukulkarni:
@andrewschoen rebased and pushed

andrewschoen:
Do we want to do this for debian systems as well?

vasukulkarni:
@andrewschoen debian is good, we dont have this issue on debian when i tested it.

PR #153, opened 2015-10-12
testnode: install QEMU packages on Fedora, RHEL, and CentOS

dillaman:
Several RBD test cases require QEMU to be present to verify integration.
Ubuntu package list already included QEMU packages.

Fixes: #13308
Signed-off-by: Jason Dillaman <dillaman@redhat.com>

PR #168, opened 2015-11-18
Revert "testnode: install QEMU packages on Fedora, RHEL, and CentOS"

vasukulkarni:
@dillaman @andrewschoen 

This has dependency on librados2 and librbd1 and endsup installing older version either on rhel or on fedora, Since @Ceph provides the right version of librados and rbd, it should be the main source.  I think it might be better to handle this at the later stage.

Check here for logs:  http://fpaste.org/291685/18351144/

Reverts ceph/ceph-cm-ansible#153

### DIFF
--- a/roles/testnode/vars/centos_6.yml
+++ b/roles/testnode/vars/centos_6.yml
@@ -81,7 +81,6 @@ packages:
   - python-matplotlib
   ###
   # for qemu
-  - qemu-kvm
   - usbredir
   - genisoimage
   ###
--- a/roles/testnode/vars/centos_7.yml
+++ b/roles/testnode/vars/centos_7.yml
@@ -66,7 +66,6 @@ packages:
   - python-matplotlib
   ###
   # for qemu
-  - qemu-kvm
   - usbredir
   - genisoimage
   ###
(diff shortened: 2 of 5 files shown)


vasukulkarni:
@dillaman is this package only required for rbd/qemu suite? we can add the ansible task there to install qemu-kvm after the install task, with this installed before install task it will have older version which is not overwritten for some distro.

dillaman:
@vasukulkarni I don't think a straight revert is the correct answer since Ubuntu does install qemu at this stage (which also will pull in librbd / librados).  You should keep the OSes consistent.

vasukulkarni:
@dillaman agree, I think it should be removed for ubuntu as well and add it after Ceph install, if you want to do it differently that is fine as well.

dillaman:
@vasukulkarni I'm fine with that approach -- but in that case, this PR should include a commit to remove the Ubuntu QEMU install and another commit for a new ansible task to install QEMU on all OSes which can be invoked via the teuthology task.

vasukulkarni:
@dillaman sure, I will include the ubuntu commit, will check with @zmc and @andrewschoen on ansible task that can used to install dependency packages for tests which can be all committed together.

wusui:
I believe that the problems in http://tracker.ceph.com/issues/13881 may be happening due to this change not being implemented.

wusui:
Note: The problem reported in my previous comment happens on trusty, so qemu-system-x86 should probably also be removed from at least the ubuntu_15.yml file as well (and probably others too).

PR #170, opened 2015-11-23
Remove bz workaround for ceph-debug

vasukulkarni:
Remove bz workaround for cephdebug-info, the corrupt yum db should be rootcaused for
proper fix

Signed-off-by: Vasu Kulkarni vasu@redhat.com

### DIFF
--- a/roles/testnode/tasks/yum/packages.yml
+++ b/roles/testnode/tasks/yum/packages.yml
@@ -10,20 +10,6 @@
   register: transaction_cleanup
   changed_when: "'Cleaning up' in transaction_cleanup.stdout"
 
-- name: Check if ceph-debuginfo is installed
-  command: rpm -q ceph-debuginfo
-  ignore_errors: yes
-  changed_when: false
-  register: bz1234967
-  tags:
-    - remove-ceph
-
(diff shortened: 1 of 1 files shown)


ktdreyer:
:+1: But the commit message could use more explanation. It explains "what", but not "why"... like why would we want to remove this workaround?

vasukulkarni:
@ktdreyer sure, I will add more info to commit message, we haven't seen this after reimage in the last few weeks, it could be due to combination of bad nodes and yum corner case issue,  I have added some more cleanup for yum systems, it should give better data in future runs.

dmick:
But...the thing is, with the workaround in place, we wouldn't have seen the bug, right?  If we didn't see it without the workaround, then the workaround would be safe to remove, but otherwise we don't know.  I vote DNM.

vasukulkarni:
@dmick with this workaround, we will still see the bug, the test will fail but this workaround helps recover the node next time by using 'rpm -e' instead of 'yum' during cleanup. so we can now push for bz to be fixed if we see it again or if its our ceph-debuginfo metadata issue it can be fixed.

PR #172, opened 2015-12-03
A new playbook to install or remove packages

andrewschoen:
See: http://tracker.ceph.com/issues/13851

### DIFF
--- a/packages.yml
+++ b/packages.yml
@@ -1,0 +1,4 @@
+---
+- hosts: all
+  roles:
+    - packages 
--- a/roles/packages/README.rst
+++ b/roles/packages/README.rst
@@ -1,0 +1,34 @@
+Packages
+========
+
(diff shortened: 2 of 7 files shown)


vasukulkarni:
LGTM, thanks @andrewschoen

zmc:
When this is run from teuthology, it looks like it will do the right thing when running in "install then remove on cleanup" mode. But will it do the right thing for "remove then reinstall on cleanup"? I can't immediately tell.

andrewschoen:
This does not support a "remove and reinstall on cleanup" feature. I could
see that working by passing a ``remove=True`` variable that would inverse
the current logic. Is that something we need for this now?

vasukulkarni:
@andrewschoen Its not required for now, we will still have to add the dependent packages to be removed in the main ansible cleanup logic at the beginning of the test.

PR #177, opened 2015-12-08
Disable lab-extras repo and iozone installs

zmc:
Tests invoking iozone will still fail. But it's better than nothing, at
least until we get the apt-mirror data we're missing.


djgalloway:
LGTM.  Easily reversible tomorrow

ktdreyer:
:+1:

iozone is non-Free software, so it's not in the distros. There was a thread a year ago about replacing it with fio, but it didn't go anywhere: https://github.com/ceph/ceph/pull/3367

dmick:
I have become convinced we should stop using iozone; it's One More Custom Thing and it's not buying us much, if anything, in test, from any discussion I can find.  Discussion here or elsewhere welcomed.

PR #178, opened 2015-12-11
Add task to partition extra drives (like on smithi)

dmick:


dmick:
corresponding variable setting on wip-nvme in ceph-sepia-secrets

PR #186, opened 2016-01-12
Revert "testnode: do not install fastcgi on RHEL"

vasukulkarni:
Sorry didn't realize without this rgw swift tests will fail on rhel, we will need this on rhel but should document for qe folks to uninstall it during upgrade tests.

Reverts ceph/ceph-cm-ansible#166

### DIFF
--- a/roles/testnode/vars/redhat_6.yml
+++ b/roles/testnode/vars/redhat_6.yml
@@ -10,7 +10,7 @@ common_yum_repos:
     priority: 2
   centos6-fcgi-ceph:
     name: "Cent OS 6 Local fastcgi Repo"
-    baseurl: http://gitbuilder.ceph.com/mod_fastcgi-rpm-rhel6-x86_64-basic/ref/master/
+    baseurl: "http://{{ gitbuilder_host }}/mod_fastcgi-rpm-rhel6-x86_64-basic/ref/master/"
     enabled: 1
     gpgcheck: 0
     priority: 2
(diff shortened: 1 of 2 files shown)


roles/testnode/vars/redhat_6.yml
```diff
@@ -8,6 +8,12 @@ common_yum_repos:
     enabled: 1
     gpgcheck: 0
     priority: 2
+  centos6-fcgi-ceph:
+    name: "Cent OS 6 Local fastcgi Repo"
+    baseurl: http://gitbuilder.ceph.com/mod_fastcgi-rpm-rhel6-x86_64-basic/ref/master/
```
andrewschoen: You should use the var ``gitbuilder_host`` here instead of hardcoding gitbuilder.ceph.com

roles/testnode/vars/redhat_7.yml
```diff
@@ -2,6 +2,12 @@
 # vars specific to any rhel 7.x version
 
 common_yum_repos:
+  rhel-7-fcgi-ceph:
+    name: "RHEL 7 Local fastcgi Repo"
+    baseurl: http://gitbuilder.ceph.com/mod_fastcgi-rpm-rhel7-x86_64-basic/ref/master/
```
andrewschoen: Same thing here, please use ``gitbuilder_host``

vasukulkarni:
@andrewschoen it now uses the var instead of the hardcode url

PR #192, opened 2016-02-03
Add vmhost role, supporting tasks, to provision VPS hosts

dmick:
This is mostly a transliteration, minus the unneeded/obsolete bits, of cookbook-vm-general.  It assumes that a host is installed and has had /srv/libvirtpool/vpmNNN set up to support the specific VMs that should live on this host.  For mira, that means the first vpm dir is on the root drive, and the 2-8th are on mounted filesystems.

Possible enhancement: make a map of vpm-to-vmhost in Ansible vars, and use it to drive the 'create /srv/libvirtpool' automatically.

### DIFF
--- a/roles/vmhost/files/libvirt-net-front.xml
+++ b/roles/vmhost/files/libvirt-net-front.xml
@@ -1,0 +1,5 @@
+<network>
+  <name>front</name>
+  <bridge name='br-front'/>
+  <forward mode='bridge'/>
+</network>
--- a/roles/vmhost/tasks/libvirt.yml
+++ b/roles/vmhost/tasks/libvirt.yml
@@ -1,0 +1,131 @@
+---
+# default pool
+- name: Query libvirt pool 'default'
+  command: virsh pool-uuid default
+  register: pool_uuid
+  failed_when: false
+
+- name: Define libvirt pool 'default'
+  command: virsh pool-define-as --name default dir --target /var/lib/libvirt/images
+  when: pool_uuid is defined and pool_uuid | failed
+
+- name: Query 'default' pool state
+  command: virsh -q pool-info default
+  ignore_errors: yes
+  register: default_pool_info
+
+- name: Start pool 'default'
+  command: virsh pool-start default
+  when: 'default_pool_info is defined and default_pool_info.stdout|search("State:  *inactive")'
+
+- name: Autostart pool 'default'
+  command: virsh pool-autostart default
+  when: 'default_pool_info is defined and default_pool_info.stdout | search("Autostart:  *no")'
+
+# Per-vpm storage pools
+
+- name: Test for /srv/libvirtpool
+  stat:
+    path: /srv/libvirtpool
+  register: srv_libvirtpool
+  failed_when: srv_libvirtpool.stat.exists is False
+
+- name: Ensure proper ownership in /srv/libvirtpool
+  file: 
+    path: /srv/libvirtpool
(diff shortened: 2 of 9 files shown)


dmick:
@djgalloway please review as well

djgalloway:
Maybe throw some of the prerequisites and quirks in a README?

I've seen the role once and I'm sure tweaks will be needed.  Can't hurt merge IMO.

roles/vmhost/files/libvirt-net-front.xml
```diff
@@ -0,0 +1,6 @@
+<network>
+  <name>front</name>
+  <!-- needs ubuntu 12.04 for libvirt >=0.9.4 -->
```
zmc: Straw man: why not just require trusty? :)

roles/vmhost/tasks/libvirt.yml
```diff
@@ -0,0 +1,130 @@
...
+  register: srv_libvirtpool
+
+- name: Ensure proper ownership in /srv/libvirtpool
+  file: 
+    path: /srv/libvirtpool
+    state: directory
+    owner: libvirt-qemu
+    group: kvm
+    recurse: yes
+  when: srv_libvirtpool.stat.exists
```
zmc: If it doesn't exist, the playbook will fail, right? Couldn't the above be tweaked to create it if necessary? You could drop the stat before this if so.

roles/vmhost/vars/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+vmhost_apt_packages:
+  - qemu-kvm
+  - libvirt-bin
+  - ethtool
+  - bridge-utils
+  #- virtinst    # ? 
+  #- ebtables    # ?
+  #- python-vm-builder # ?
```
zmc: Are these commented because you're unsure if they're necessary?

dmick: Yeah, good point.  I'll clean those up.  (I'm also not 100% sure about ethtool)

roles/vmhost/tasks/libvirt.yml
dmick: Yeah, it's probably better that to just fail the play if /srv/libvirtpool doesn't exist; nothing good will happen.

roles/vmhost/files/libvirt-net-front.xml
dmick: I'll remove the comment.

dmick: and yeah, a readme makes sense.

roles/vmhost/tasks/libvirt.yml
zmc: What I was getting at, though, was 'just create it' - is that not possible?

dmick: In theory yes, but it's a lot more work; see the readme.  It needs to be created, have subdirs, have filesystems on disks and mounts of those disks on the subdirs; to know what those subdirs should be called is a new set of information mapping vpms to vmhosts.  It's not *that* much work, but it's a lot more conditional/error checking, and I'd want to do it as a followon to this.

zmc: Understood. Thanks!

zmc:
:+1:

djgalloway:
LGTM

dmick:
merge at will.

PR #195, opened 2016-02-15
Stop looking for keys on git.ceph.com

dmick:


dmick:
(there are corresponding changes to sepia- and octo-secrets)

ktdreyer:
LGTM

PR #197, opened 2016-02-16
Enable Fedora 22 installation in labs

djgalloway:
http://tracker.ceph.com/issues/14729


roles/testnode/tasks/pip.yml
```diff
@@ -1,16 +1,18 @@
 ---
+# python-pip is installed from epel repo which doesn't exist on fedora.
+# package is installed during kickstart instead.
```
dmick: Do you mean "package is part of default installation on Fedora" here?  (that is, please change it to something like that)

roles/testnode/vars/fedora_22.yml
```diff
@@ -70,28 +61,20 @@ packages:
...
   # for disk/etc monitoring
+  - nrpe
+  - nagios-plugins-all
   - smartmontools
   # for nfs
   - nfs-utils
+  # needed until we stop using yum ansible module
+  - yum
+  # python-pip is installed via roles/testnode/tasks/pip.yml on other rpm-based distros
+  - python-pip
```
dmick: but wait, why is this still here if it was "installed during kickstart"?...

djgalloway: Because these packages get installed when the post-install trigger is called and cobbler runs the testnodes role on the host.

roles/testnode/tasks/pip.yml
djgalloway: python-pip is not part of default installation on Fedora.  It gets installed during/after a kickstart when cobbler runs ansible on the host.

roles/testnode/vars/fedora_22.yml
dmick: right, but, if it's already installed, ...

roles/testnode/tasks/pip.yml
dmick: OK.  So it's not part of kickstart; it's part of the post-install ansible.  So how about "package is installed during post-install ansible (or name the task specifically, or both)

dmick:
LGTM.  @zmc, you done?

zmc:
:beers:

PR #202, opened 2016-02-25
Added "- ncurses-devel" for xfstests to run and pass

yuriw:
Fixes http://tracker.ceph.com/issues/14848

Signed-off-by: Yuri Weinstein <yweinste@redhat.com>

### DIFF
--- a/roles/testnode/vars/centos_7.yml
+++ b/roles/testnode/vars/centos_7.yml
@@ -87,6 +87,8 @@ packages:
   - smartmontools
   # for nfs
   - nfs-utils
+  # for xfstests
+  - ncurses-devel
 
 epel_packages:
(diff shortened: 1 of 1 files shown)


ktdreyer:
It would be great if the particular *test* that need this package could add it instead

ktdreyer:
Sigh, nevermind, I see that was already shot down in the Redmine ticket.

dmick:
Yeah, I'd rather have it there, but this is not the time I suppose.  The thing is we'll never go back and clean these up.  

oh well.

dmick:
yuriw: add a line to the commit message that says, exactly:

Fixes: #14848

just above your Signed-off-by line.

PR #203, opened 2016-02-28
ansible.cfg: do not set vault_password_file

dachary:
So that ansible-playbook can run from the root of the repository and not
pick up this value which is unlikely to be correct.

http://tracker.ceph.com/issues/14914 Fixes: #14914

Signed-off-by: Loic Dachary <loic@dachary.org>


dachary:
Hum, this is not the right fix. I thought the file had to exist on the target but it needs to exist on the host.

PR #209, opened 2016-03-15
testnodes: Disable ssl requirement for nrpe on rpm-based distros

djgalloway:
The variable name the systemd nrpe service looks for is <code>NRPE_SSL_OPT</code>
See <code>/usr/lib/systemd/system/nrpe.service</code>
The variable name nrpe on Ubuntu looks for is <code>DAEMON_OPTS</code>

Since the variable name *and* default configuration location differ between distros, we can get rid of the <code>roles/testnode/templates/nagios/nagios-nrpe-server</code> template and just set the single variable instructing NRPE to not use SSL.

### DIFF
--- a/roles/testnode/tasks/nagios.yml
+++ b/roles/testnode/tasks/nagios.yml
@@ -8,13 +8,19 @@
     mode: 0440
     validate: visudo -cf %s
 
-- name: Configure nagios nrpe settings.
-  template:
-    src: nagios/nagios-nrpe-server
-    dest: "/etc/default/{{ nrpe_service_name }}"
-    owner: root
-    group: root
-    mode: 0644
+- name: Configure nagios nrpe settings (Ubuntu)
+  lineinfile:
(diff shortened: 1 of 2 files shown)


roles/testnode/tasks/nagios.yml
```diff
@@ -8,13 +8,19 @@
...
-  template:
-    src: nagios/nagios-nrpe-server
-    dest: "/etc/default/{{ nrpe_service_name }}"
-    owner: root
-    group: root
-    mode: 0644
+- name: Configure nagios nrpe settings (Ubuntu)
+  lineinfile:
+    dest: /etc/default/{{ nrpe_service_name }}
+    regexp: "^DAEMON_OPTS.*"
```
dmick: the .* isn't necessary, although it doesn't really hurt anything; it only has to match something on the line, not the whole line

djgalloway: Good call.  Verified and updated.

PR #212, opened 2016-03-16
testnodes: configure selinux for nagios monitoring

djgalloway:
This addresses SELinux denials on CentOS and RHEL testnodes.

### DIFF
--- a/roles/testnode/files/nagios/nrpe.te
+++ b/roles/testnode/files/nagios/nrpe.te
@@ -1,0 +1,12 @@
+module nrpe 1.0;
+
+require {
+	type fsadm_exec_t;
+	type nrpe_t;
+	type hwdata_t;
+	class file { read getattr open };
+}
+
+#============= nrpe_t ==============
+allow nrpe_t fsadm_exec_t:file getattr;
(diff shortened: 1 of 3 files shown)


djgalloway:
@zmc What do you think about this approach?

I've tested on mira101.  The output makes it back to nagios but isn't correct.  I've confirmed that issue's not SELinux related, however, by disabling SELinux on mira101 as a sanity check.

zmc:
@djgalloway Looks pretty good! What's wrong with the output? Should it be tested on a smithi?

djgalloway:
@zmc I just tested it on smithi019 and output is as expected.  The mangled output on mira101 could be anything.. May need a reimage, may just need the whole testnodes playbook re-run but I'm pretty confident it's not SELinux.
<pre>
#  Output on mira101
0 of drives failing/missing
Controller needs newer firmware for S.M.A.R.T. support
</pre>
Also found another package, <code>libsemanage-python</code> was required to use the seboolean ansible module so threw that in as an additional task in my latest push.

zmc:
Awesome, thanks @djgalloway

PR #216, opened 2016-03-22
gateway: Adding additional functionality and some missing files

djgalloway:


roles/gateway/tasks/users.yml
```diff
@@ -11,3 +11,11 @@
     owner: root
     group: root
     mode: 0644
+
+- name: Upload auth-openvpn script
+  template:
```
dmick: This doesn't need to be a template, right?  there's no substitutions I see?

roles/gateway/tasks/main.yml
```diff
@@ -19,6 +19,7 @@
 # Manage OpenVPN users list using secrets repo
 - include: users.yml
   tags:
+    - always
```
dmick: Why always?  and if we need it, doesn't this make -users pointless?

roles/gateway/README.rst
```diff
@@ -23,11 +23,15 @@ A list of packages to install that is specific to the role.  The list is defined
 
     packages: []
 
+A unique name to give to your OpenVPN service.  This name is used to organize configuration files and start/stop the service.  Defined in the secrets repo::
+
+    openvpn_server_name: []
```
dmick: If this isn't given a default, is there something which will check for it being set and error if not?

roles/gateway/tasks/main.yml
djgalloway: I feel like I had a better explanation for this yesterday but now all I can think of is "why not?"

One can still call the users tag directly and only run those tasks so the tag isn't useless.

roles/gateway/README.rst
djgalloway: Good point.  Added https://github.com/ceph/ceph-cm-ansible/pull/216/commits/1cf6fd5b68de0f5c21c7f1367b84d2c94811b5fe

roles/gateway/tasks/main.yml
dmick: as to "why not": what if I want to set up an openvpn server but not set up the users file?

roles/gateway/defaults/main.yml
```diff
@@ -5,3 +5,7 @@
 secrets_repo:
   name: UNDEFINED
   url: null
+
+openvpn_server_name: "server"
```
dmick: I know you don't need quotes here; not sure if they do any damage, but it's not the prevailing style, so, should remove them

roles/gateway/tasks/main.yml
djgalloway: `--skip-tags="users"` is a thing, but point taken.  Removed.

PR #227, opened 2016-04-21
Add pcp-webapi on ubuntu upstream setups; set pmcd connect timeout for pmmgr

zmc:


roles/pcp/tasks/web.yml
```diff
@@ -1,8 +1,8 @@
 ---
 - name: Fail when on Ubuntu
   fail:
-    msg: "pcp-web isn't available on Ubuntu yet :("
-  when: ansible_distribution == "Ubuntu"
+    msg: "pcp-webapi is only available when using upstream packages. Set upstream_repo to true."
+  when: ansible_distribution == "Ubuntu" and upstream_repo|bool != true
```
djgalloway: Might be time to start double-quoting variables in new/modified code since Ansible v2 throws a warning for each unquoted variable.

zmc: Ok, this is strange: I can't get that deprecation warning to display. If you give me the line I should be using, I can repush this with it.

djgalloway: Nevermind.  Looks like bare variables are okay in conditionals.

PR #228, opened 2016-04-21
common: Replace bad epel mirror

djgalloway:
Mirror fedora-epel.mirror.lstn.net is almost a week out of date and
causing yum package transactions to fail due to bad repodata.  Replacing
with a mirror closer to the sepia and octo labs.

Fixes: http://tracker.ceph.com/issues/15539

### DIFF
--- a/roles/common/templates/epel-mirrorlist
+++ b/roles/common/templates/epel-mirrorlist
@@ -3,5 +3,5 @@
 # local yum mirrorlist for epel-{{ ansible_distribution_major_version }}
 http://mirrors.cat.pdx.edu/epel/{{ ansible_distribution_major_version }}/$basearch
 http://mirror.pnl.gov/epel/{{ ansible_distribution_major_version }}/$basearch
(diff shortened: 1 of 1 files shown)


yuriw:
@djgalloway @zmc  - can we build some protection  "a la" compare mirrors dirs and if not equal fail ?

djgalloway:
@dmick and I were discussing this earlier.

My theory is perhaps this all started due to DNS failures before we knew DNS was out to get us so we kept adding and tweaking unnecessary workarounds.

Having a local mirrorlist we have to maintain obviously isn't ideal but I don't think it's worth putting any effort into comparing mirrors when our current mirrorlist method is already jury-rigged.

zmc:
@yuriw Interesting idea, but I'm not sure if the work it'd require would be much less than the time it could potentially save us

dmick:
Still pissed off that yum/dnf don't *tell me what's wrong*.

I think a better middle ground between "we have a static list of mirrors" and "we download the mirrorlist everytime and thus fail often" would be to cron-download the mirror list periodically (say, to apt-mirror) and configure to use the local copy

PR #229, opened 2016-04-21
A role for an internal master nameserver

djgalloway:
Will work with changes to secrets repo once https://github.com/ceph/ceph-sepia-secrets/pull/103 is merged.


roles/nameserver/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+# These defaults are present to allow certain tasks to no-op if a secrets repo
+# hasn't been defined. If you want to override these, do so in the secrets repo
+# itself. We override these in  $repo/ansible/inventory/group_vars/nameserver.yml
+secrets_repo:
+  name: UNDEFINED
```
zmc: Why not null here as well?

djgalloway: No specific reason.  I took it from https://github.com/ceph/ceph-cm-ansible/blob/master/roles/cobbler/defaults/main.yml#L6 originally and used it in the gateway role as well but I see where you actually check for "UNDEFINED" in the cobbler role.

It may not be needed at all since the play will bail out if there's no secrets repo anyway.

roles/nameserver/tasks/config.yml
```diff
@@ -0,0 +1,22 @@
...
+    src: named.conf.j2
+    dest: "{{ named_conf_file }}"
+    validate: named-checkconf %s
+  notify: restart named
+  with_dict: "{{ named_domains }}"
+
+- name: Apply named daemon options
+  lineinfile:
+    dest: /etc/sysconfig/named
+    regexp: '^'
```
zmc: This looks like it will match any line? Do you want ``^OPTIONS=`` ?

zmc:
Excellent work @djgalloway. As far as I'm concerned this can ship as-is. I imagine @dmick will catch something I missed though :)

roles/nameserver/tasks/config.yml
djgalloway: Yep. I changed the way I wanted this to work in a recent change and forgot to re-add `OPTIONS`.  Will fix.

roles/nameserver/tasks/firewall.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Enable firewalld
```
dmick: Is starting firewalld enough to take over from any other services, like a default iptables or whatever?

roles/nameserver/README.rst
```diff
@@ -0,0 +1,170 @@
...
+Prerequisites
++++++++++++++
+
+- CentOS 7.2
+
+Variables
++++++++++
+Most variables are defined in ``roles/nameserver/defaults/main.yml`` and values are chosen to support our Sepia_ lab.  They can be overridden in the ``secrets`` repo.
+
+A list of packages to install that is specific to the role.  The list is defined in ``roles/nameserver/vars/packages.yml``::
```
dmick: I'm confused by this style of "define the variable, and then show it"; it may be what other doc files have done, but I think it'd be way better to say 

packages: []

A list of packages to install....

roles/nameserver/README.rst
```diff
@@ -0,0 +1,170 @@
...
+named daemon options.  Writes to ``/etc/sysconfig/named``.  Defined in ``roles/nameserver/defaults/main.yml``::
+
+    named_conf_daemon_opts: []
+
+    # Example for IPv4 support only:
+    named_conf_daemon_opts: "-4"
+
+The following variables are self-explanatory and written to each zone file's SOA resource record.  They're defined in ``roles/nameserver/defaults/main.yml``::
+
+    named_conf_ttl: 3600
```
dmick: It might be slightly clearer to name these "named_conf_soa_ttl", just to make it clear in the code and in the variable file that they are for the SOA record only.  The doc couldn't be clearer, but for those times you're editing and aren't looking at the doc...

roles/nameserver/README.rst
```diff
@@ -0,0 +1,170 @@
...
+The ``named_domains`` dictionary is the bread and butter of creating zone files.  It is in standard YAML syntax.  Each domain (key) must have ``forward`` and ``ipvar`` defined although ``ipvar`` can be set to ``NULL``.  Optional values include ``miscrecords`` and ``reverse``.
+
+``forward``
+  The domain of the forward lookup zone for each domain (key)
+
+``ipvar``
+  The variable assigned to a system in the Ansible inventory.  This allows systems to have multiple IPs assigned for a front and ipmi network, for example.  See **Inventory Example** below.
+
+``miscrecords``
+  Records to add to corresponding ``forward`` zone file.  This is a good place for CNAMEs and MX records and records for hosts you don't have in your Ansible inventory.  If your main nameserver is in a subdomain, you should create a record here.  See example.
```
dmick: "If your main nameserver is in a subdomain, you should create a record here"...does this mean the glue record, or something else?

dmick:
Gotta agree that overall this looks amazing.  The commit organization is perfect, and things are just about as clear as they can be (man, that with_subelements stuff is gnarly).

Mostly doc comments.  Nice job.

roles/nameserver/tasks/firewall.yml
djgalloway: firewalld is the default firewall in CentOS 7.  The iptables service isn't installed by default anymore.

roles/nameserver/README.rst
djgalloway: It's laid out this way in every other README with variables.  How about a table instead?

roles/nameserver/README.rst
djgalloway: Yup

roles/nameserver/README.rst
djgalloway: K

PR #230, opened 2016-04-22
pcp: Correct /var/log/pcp permissions for collectors

zmc:
http://tracker.ceph.com/issues/15575

Some hosts were getting a /var/log/pcp/pmmgr that was owned by root:root
for some reason.


djgalloway:
Can you update the commit message to have `Fixes: http://tracker.ceph.com/issues/15575` at the bottom?  I guess that's the standard now.

zmc:
@djgalloway done!

PR #231, opened 2016-04-22
testnode: Update SELinux policy when configuring NRPE

djgalloway:
The 'nagios_run_sudo' boolean was added after RHEL7.1 GA.

Attempting to set the boolean when it's not present causes ansible
to fail.  NRPE will cause SELinux failures in test jobs unless set
so the only other option would be to disable NRPE on
RHEL/CentOS <= 7.1

This change ensures the testnode has the latest version of SELinux
policy installed, and thus, ensuring the boolean exists.


vasukulkarni:
:+1:

PR #233, opened 2016-04-27
cobbler: Only set bootdev in kickstart if installing Xenial.

djgalloway:
"bootdev string default" is only supported in Xenial apparently.

Discovered this after attempting to install Trusty using the new kickstart.

### DIFF
--- a/roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
+++ b/roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
@@ -110,7 +110,9 @@ d-i pkgsel/upgrade select safe-upgrade
 d-i pkgsel/update-policy select none
 
 # Install GRUB to default location.  Required as of Xenial.
+#if $os_version == 'xenial'
 d-i grub-installer/bootdev  string default
+#end if
 
 # During installations from serial console, the regular virtual consoles
 # (VT1-VT6) are normally disabled in /etc/inittab. Uncomment the next


roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
```diff
@@ -110,7 +110,9 @@ d-i pkgsel/upgrade select safe-upgrade
 d-i pkgsel/update-policy select none
 
 # Install GRUB to default location.  Required as of Xenial.
+#if $os_version == 'xenial'
```
dmick: no >= of some numeric available?

djgalloway: I mimicked the existing `if` statements but you're right.. this particular check should probably be based on a numerical major version to future-proof it.

dmick: nah.  just file a followup (presuming you have a major or a 'major*100+minor" number you can use)

zmc: Good point @dmick - looks like I was a bit hasty :)

djgalloway: Fixed in https://github.com/ceph/ceph-cm-ansible/pull/234

PR #234, opened 2016-04-27
cobbler: Only set bootdev in kickstart if installing Xenial.

djgalloway:
"bootdev string default" is only supported in Xenial.

This is a saner/future-proof method of implementing https://github.com/ceph/ceph-cm-ansible/pull/233

### DIFF
--- a/roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
+++ b/roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
@@ -3,6 +3,12 @@
 # Fetch the os_version from the distro using this profile.
 #set os_version = $getVar('os_version','')
 
+# Fetch Ubuntu version (e.g., 14.04)
+#set distro_ver = $getVar('distro','').split("-")[1]
+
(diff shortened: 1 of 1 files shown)


roles/cobbler/templates/kickstarts/cephlab_ubuntu.preseed
```diff
@@ -3,6 +3,12 @@
 # Fetch the os_version from the distro using this profile.
 #set os_version = $getVar('os_version','')
 
+# Fetch Ubuntu version (e.g., 14.04)
+#set distro = $getVar('distro','').split("-")[1]
+
```
dmick: Is the variable namespace where getVar('distro') is defined different from the namespace where set defines it (i.e. are we redefining distro)?  If so, maybe we should choose a different local name in case something else depends on the current value

djgalloway: Good call.  `distro` is already a variable defined in each profile.

I changed `distro` to `distro_ver` and `distro_ver` to `distro_ver_major`.

PR #235, opened 2016-04-27
testnode: Remove manual configuration of getty

djgalloway:
GRUB2 handles setting up getty during installation so the tasks
and templates removed in this commit are no longer needed.

The main reason attention was brought to them was because the getty service name changed (`s/ttyS1/getty@ttyS1.service`) with the introduction of systemd to Ubuntu.  The service handler would fail causing ansible to fatally exit.


dmick:
It's not Grub2, but it is automatic.

PR #236, opened 2016-04-28
cobbler: Prevent systemd from renaming NICs

djgalloway:
systemd in Xenial was renaming NICs to `eno1`.

Since we ship a static `/etc/network/interfaces` file with `eth0` as the only NIC name, networking wouldn't come up on boot.

Tested on clara and magna node.


ktdreyer:
This feels like one of those things where we are changing the labs to make things easier, while moving further away from what users and customers will do on their systems.

PR #239, opened 2016-05-05
pcp: retention settings for managers

zmc:


roles/pcp/README.rst
```diff
@@ -53,6 +53,10 @@ connect to hosts that it monitors (in seconds)::
 
     pmcd_connect_timeout: 1
 
+To tell a ``pcp_manager`` host to retain full-resolution archives for a year::
```
dmick: can we have a reference to the doc that describes legal forms of this param (the example makes it clear that "Ndays" is valid, but can I say hours, or weeks, or...)

roles/pcp/defaults/main.yml
```diff
@@ -17,6 +17,8 @@ pcp_probe: false
 pmcd_connect_timeout: "0.1"
 # How long to keep full-resolution archives before reducing to save space
 pmlogmerge_retain: "90days"
+# How long before deleting reduced archives
+pmlogmerge_reduce: "99999weeks"
```
dmick: oh look apparently we can say weeks

zmc:
@dmick - forgot to update this one, but I've repushed with links to docs

PR #241, opened 2016-05-11
pcp: update upstream packages

zmc:
PCP's upstream doesn't version its dependencies in all cases. This leads
to situations where, when setting upstream_repo to True, a system ends
up with some old and some new packages. This results in a potentially
broken installation.

This commit causes all installed pcp packages to be upgraded when
upstream_repo is True.


dmick:
This would have been a *whole* lot less confusing if you'd mentioned "since repo.yml isn't included when upstream_repo is false, do the apt update in main.yml".  That took a long while to untangle in my head because I was looking for behavior based on upstream_repo == True.

zmc:
sorry @dmick :(

dmick:
Not that big of a deal, just setting levels about context :)

djgalloway:
Context?! :smile:

dmick:
Oh you!  (Insert meme here)

PR #243, opened 2016-05-17
testnode/tasks/apt/packages.yml: just force everything

athanatos:
Signed-off-by: Samuel Just <sjust@redhat.com>

### DIFF
--- a/roles/testnode/tasks/apt/packages.yml
+++ b/roles/testnode/tasks/apt/packages.yml
@@ -3,6 +3,7 @@
   apt:
     name: "{{ item }}"
     state: absent
+    force: yes
   with_items: ceph_packages_to_remove|list + packages_to_remove|list
   when: ceph_packages_to_remove|length > 0 or
         packages_to_remove|length > 0
@@ -29,6 +30,7 @@
   apt:
     name: "{{ item }}"
     state: present 
+    force: yes
   with_items: packages|list + common_packages|list
   when: packages|length > 0 or
         common_packages|length > 0
@@ -37,5 +39,6 @@
   apt:
     name: "{{ item }}"
     state: present
+    force: yes
   with_items: non_aarch64_packages|list + non_aarch64_common_packages|list
   when: ansible_architecture != "aarch64"


athanatos:
Tested: sjust@teuthology:/a/samuelj-2016-05-16_20:02:21-rados-wip-sam-testing-distro-basic-smithi/197087

zmc:
@athanatos this makes me a little nervous. Some justification for the change would help me understand why it's necessary. Thanks!

athanatos:
@zmc Same reason I added force elsewhere, the install failed with a message indicating that -y was used without --force.  I didn't diagnose further.

zmc:
@athanatos that's not ringing a bell. Was it recent?

zmc:
```
[15:25:40]  <zackc>	sjusthm: ok. so, i don't have a memory of the "same reason"
[15:26:05]  <sjusthm>	zackc: let me try to find the run
[15:29:51]  <zackc>	sjusthm: is it this pr? https://github.com/ceph/ceph-cm-ansible/pull/221
[15:30:13]  <sjusthm>	yes
[15:30:17]  <sjusthm>	found an example
[15:30:23]  <sjusthm>	sjust@teuthology:/a/samuelj-2016-05-10_19:44:43-rados-wip-split-lfn-hammer-distro-basic-smithi/183862
[15:31:53]  <zackc>	oh, package auth problem
[15:32:00]  <zackc>	was that ever root-caused?
[15:32:08]  <zackc>	it's kind of alarming
[15:32:16]  <zackc>	i vaguely remember looking into it
[15:32:28]  <sjusthm>	no, it was never root caused
[15:32:35]  <sjusthm>	I mostly just want to avoid having it mess up the tests
[15:32:50]  <sjusthm>	no reason it can't be root caused in parallel
[15:32:56]  <sjusthm>	but at least the tests will fail less often
[15:34:22]  <zackc>	yeah, i suppose you're right
```

There's no reason to block this particular PR, even though our use of ``force`` everywhere bugs me. So I'll merge :)

PR #245, opened 2016-05-31
testnode: Replace old smart.pl with rewritten smart.sh script

djgalloway:
Tested on:
- senta01 (ARC-1222)
- senta02 (mdadm RAID)
- mira058 (ARC-1222 with one failing drive)
- mira124 (ARC-1222 with 5 bad and 1 missing disk)
- smithi005 (No bad disks)
- magna008 (1 failing disk)

Fixes: http://tracker.ceph.com/issues/14682


roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+### Pre-flight checks ###
+# Set raid var then check for cli64 command and bail if missing
+if lspci | grep -qi areca
+then
+  raid=true
+else
+  raid=false
+fi
+
+if ! [ -x "$(command -v cli64)" ] && [ "$raid" = true ]
```
dmick: If you swap the order, then the shell doesn't have to hunt for the command if it's not doing raid.  Somewhat more efficient as well as more logical

djgalloway: Smart.  Thanks for the tip

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+then
+  echo "ERROR - cli64 command not found or is not executable"
```
dmick: Some might say that error messages should go to stderr (add >&2)

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+# Check for smartmontools and bail if missing
+if ! [ -x "$(command -v smartctl)" ]
+then
+  echo "ERROR - smartctl is not installed or is not executable"
+  echo "yum/apt-get install smartmontools"
+  exit 3
+fi
+
+### Variables ###
+numdrives=0 # total number of drives (or RAID slots) discovered
```
dmick: style: in general, comments to the right make things harder to read.  If one must, keep 'em short and aligned; otherwise 'line above the variable' is preferable

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+
+### Functions ###
+
+# Gather smart data for drives behind Areca RAID controller
+# NOTE: Only works when individual disks set to JBOD for now
+areca_smart ()
+{
+  numdrives=$(sudo cli64 disk info | grep -E "Slot#[[:digit:]]" | wc -l)
+  # Loop through all disks not marked as 'N.A.' or 'Failed'
+  for slot in $(sudo cli64 disk info | grep -E "Slot#[[:digit:]]" \
```
dmick: since this is the same pipeline as above, you could execute it once and save its result, and gain some efficiency and some readability.  Then you could both wc -l it and use it to feed the grep/cut here

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+    failed=false
+    output=$(sudo cli64 disk smart drv=$slot \
+    | grep -E "^  "5"|^"197"|^"198"" | awk '{ print $(NF-1) }')
+    outputcount=$(echo $output | wc -w)
+    # Determine drive letter given slot number
+    dl=$(areca_bay_to_letter $slot)
+    # Only continue if we received 3 SMART data points
+    if [ "$outputcount" = "3" ]
+    then
+      read reallocated pending uncorrect <<< $output
```
dmick: I literally had never seen this construct before in my life.

roles/testnode/files/libexec/smart.sh
djgalloway: Any echoed message is intended to appear on the main Nagios screen.  Does that matter?

dmick: Ah.  I dunno.  maybe.  I guess it depends on what Nagios does with stderr vs stdout

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+        failed=true
+        rc=2
+      fi
+      if [ "$uncorrect" != "0" ]
+      then
+        messages+=("Drive $slot $dl has $uncorrect uncorrect sectors")
+        failed=true
+        rc=2
+      fi
+    elif [ "$outputcount" != "3" ]
```
dmick: Is this the same as "else"?  (the if was == 3)

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+  | grep -v 'N.A.\|Failed' | grep -o "Slot#[[:digit:]]" | cut -c6-)
+  do
+    failed=false
+    output=$(sudo cli64 disk smart drv=$slot \
+    | grep -E "^  "5"|^"197"|^"198"" | awk '{ print $(NF-1) }')
+    outputcount=$(echo $output | wc -w)
+    # Determine drive letter given slot number
+    dl=$(areca_bay_to_letter $slot)
+    # Only continue if we received 3 SMART data points
+    if [ "$outputcount" = "3" ]
```
dmick: if this long if/then gets unwieldy, you can compress it a little with if <condition>; then  (move the then up a line) without losing much readability.

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+    # Make sure drives with multiple types of bad sectors only get counted once
+    if [ "$failed" = true ]
+    then
+      let "failingdrives+=1"
+    fi
+  done
+}
+
+# Correlate Areca drive bay to drive letter
+areca_bay_to_letter ()
```
dmick: depending how often this is called, it might be worth caching, or 'memoizing' in the function here (basically: a local cache that it checks first, and just returns the value if it has it)  Obviously this works, but as a future optimization in case you want it, that might be low-hanging fruit.

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+    # If cli64 and smartctl find a S/N match, return drive letter
+    if [ "$areca_serial" = "$smart_serial" ]
+    then
+      echo "($dl)"
+    fi
+  done
+}
+
+# Standard SATA/SAS drive smartctl check
+normal_smart ()
```
dmick: It seems like this shares a lot of code with the areca version; would it be profitable to try to factor out the common stuff into functions?..

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,254 @@
...
+  | grep 'Failed' | awk '{ print $1 }')
+  do
+    messages+=("Drive $drive failed")
+    let "failingdrives+=1"
+    rc=2
+  done
+fi
+
+## Process all other drives
+if [ "$raid" = false ]
```
dmick: else?

roles/testnode/files/libexec/smart.sh
djgalloway: K. Fixed in latest revision.

roles/testnode/files/libexec/smart.sh
djgalloway: Yeah I knew I was unnecessarily calling cli64 more than needed.  Just figured this trick out...

```
# output=$(sudo cli64 disk info)                                                                  
# 
# echo $output
# Enc# Slot# ModelName Capacity Usage =============================================================================== 1 01 Slot#1 ST31500341AS 1500.3GB Raid Set # 000 2 01 Slot#2 ST31500341AS 1500.3GB Raid Set # 000 3 01 Slot#3 ST1500DM003-9YN16G 1500.3GB Raid Set # 000 4 01 Slot#4 ST31500341AS 1500.3GB Raid Set # 000 5 01 Slot#5 ST1500DM003-9YN16G 1500.3GB Free 6 01 Slot#6 ST31500341AS 1500.3GB Raid Set # 000 7 01 Slot#7 N.A. 0.0GB N.A. 8 01 Slot#8 ST1500DM003-9YN16G 1500.3GB Raid Set # 000 =============================================================================== GuiErrMsg<0x00>: Success.
# 
# 
# echo "$output"
  # Enc# Slot#   ModelName                        Capacity  Usage
===============================================================================
  1  01  Slot#1  ST31500341AS                     1500.3GB  Raid Set # 000  
  2  01  Slot#2  ST31500341AS                     1500.3GB  Raid Set # 000  
  3  01  Slot#3  ST1500DM003-9YN16G               1500.3GB  Raid Set # 000  
  4  01  Slot#4  ST31500341AS                     1500.3GB  Raid Set # 000  
  5  01  Slot#5  ST1500DM003-9YN16G               1500.3GB  Free      
  6  01  Slot#6  ST31500341AS                     1500.3GB  Raid Set # 000  
  7  01  Slot#7  N.A.                                0.0GB  N.A.      
  8  01  Slot#8  ST1500DM003-9YN16G               1500.3GB  Raid Set # 000  
===============================================================================
GuiErrMsg<0x00>: Success.
```

Fixed in latest revision

roles/testnode/files/libexec/smart.sh
djgalloway: Not the same as `else`.  I think at one point, I had more than one conditional I was checking but `if .. then .. else` should work fine here now.

roles/testnode/files/libexec/smart.sh
djgalloway: Eh, it only gets called if a drive has any bad sectors in the latest version I've yet to push.  Once we have all our bad drives replaced (haha) it shouldn't get called all that often.

roles/testnode/files/libexec/smart.sh
djgalloway: I thought about this and you're right.

The main reason I've kept them separate so far is in case we have to go back to the legacy way of polling SMART data from disks behind an Areca controller.  (`sudo smartctl -a -d areca,$drive $scsidev`)

I also need to tweak `areca_smart ()` a bit more so it can handle disks in part of a hardware RAID.

roles/testnode/files/libexec/smart.sh
djgalloway: I split counting the counting of N.A. and failed drives into a separate function and created a main function that checks if raid=true,false, or not set at all.

djgalloway:
@dmick New version pushed that addresses most of your comments.

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,284 @@
...
+    exit 3
+  fi
+}
+
+# Gather smart data for drives behind Areca RAID controller
+areca_smart ()
+{
+  # Store output of cli64 to reduce repeated executions
+  cli64out=$(sudo cli64 disk info | grep -E "Slot#[[:digit:]]")
+  numdrives=$(echo "$cli64out" | wc -l)
```
dmick: now that I know about <<<, it would work here too.  six of one, since echo is a shell builtin anyway.

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,284 @@
...
+    if [ "$failed" = true ]
+    then
+      let "failingdrives+=1"
+    fi
+  done
+}
+
+## Call main() function
+main
+
```
dmick: it's kind of weird to have a "main" function that's not "everything in the script that's not a function" (as main traditionally is in C, Python, etc.)  I'd either subsume the rest of the code into main, or rename main.

roles/testnode/files/libexec/smart.sh
djgalloway: Not sure I follow your suggestion here

roles/testnode/files/libexec/smart.sh
djgalloway: So @zmc suggested I create a main() function.  Should all of [this](https://github.com/ceph/ceph-cm-ansible/blob/9c8cee1999d82a0441564cfa197c4a8a27a66321/roles/testnode/files/libexec/smart.sh#L251-L284) go in it as well or am I missing the point of having a main function?

roles/testnode/files/libexec/smart.sh
dmick: Really just musing that it could be numdrives=$(wc -l <<<"$cli64out")....although maybe it couldn't because maybe that would join lines.  anyway, ignore it.

roles/testnode/files/libexec/smart.sh
dmick: my inclination is that all of that should go into it *and* to not is missing the point of a main function, yes.

djgalloway: Fixed!

roles/testnode/files/libexec/smart.sh
```diff
@@ -0,0 +1,285 @@
...
+# total number of drives (or RAID slots) discovered
+numdrives=0
+# Number of failed, failing, and/or missing drives
+failingdrives=0
+# Fallback message for UNKNOWN return code output
+unknownmsg="Unknown error"
+# Return code for nagios (Default to UNKNOWN)
+rc=3
+# Array of messages indicating drive health.  Output after nagios status.
+declare -a messages
```
zmc: Could these lines above go in ``main()``?

djgalloway: I tried that but then the areca functions couldn't set or read them.  Didn't do a ton of research into why.

dmick: that should only be the declare, and you can make it global with -g I think.  experimenting

djgalloway:
@zmc I've moved preflight checks to a separate function that's called by main.

dmick:
Looks real good, and I know zmc ran out of time, so I'll merge

PR #246, opened 2016-06-04
smart.sh: default rc to 0 (saw false UNKNOWNs with non-raid)

dmick:


dmick:
Merging my own PR to stop nagios noise over the weekend.

PR #249, opened 2016-06-17
Network configuration for gateway role

djgalloway:
Tested on fresh CentOS 7.2 VM in RHEV.  To test the role and avoid IP conflicts, check out the wip-gw-networking branch from the secrets repo and change the following:

- In `ansible/inventory/sepia`, replace gw.sepia.ceph.com with the VM's internal IP under `[gateway]`
- In `ansible/secrets/gateway.yml`, change the `local` line in the OpenVPN server config to a free public IP (I used 8.43.84.133)
- In `ansible/secrets/gateway.yml`, change the `server` line in the OpenVPN server config to something like `server 172.21.50.0 255.255.255.0`

A RHEV VM is the only place the role can really be tested since no other machines are configured to be reachable from the WAN.

This should *not* be run on the existing OpenVPN server.

21JUN2016 EDIT: Now that we've moved the gw into RHEV, the role can be tested using public IP 8.43.84.134.  The role can also safely be run on the existing gw VM (172.21.0.100).

### DIFF
--- a/roles/gateway/README.rst
+++ b/roles/gateway/README.rst
@@ -152,6 +152,5 @@ To Do
 
 - Support installation of new OpenVPN gateway from scratch
 - Generate and pull (to secrets?) CA, keys, and certificates
-- Configure log rotation
 
 .. _Sepia: https://ceph.github.io/sepia/
--- a/roles/gateway/README.rst
+++ b/roles/gateway/README.rst
@@ -8,6 +8,9 @@ This role supports CentOS 7.2 only at this time.  Its current intended use
 is to maintain the existing OpenVPN gateway in our Sepia_ lab.
 
 It does the following:
+- Configures network devices
+- Configures firewalld
+- Configures fail2ban
 - Installs and updates necessary packages
 - Maintains user list
 
@@ -58,6 +61,64 @@ A list of users that don't have their ssh pubkey added to the ``teuthology_user`
     openvpn_users:
       - ovpn: user@host etc...
 
+The following vars are used to populate ``/etc/resolv.conf``.  Defined in the
+secrets repo::
+
+    gw_resolv_search: []
+    # Example: gw_resolv_search: "front.example.com"
+
+    gw_resolv_ns: []
+    # Example:
+    gw_resolv_ns:
+      - 1.2.3.4
+      - 8.8.8.8
+
+The ``gw_networks`` dictionary assumes you have individual NICs for each
+VLAN in your lab.  The subelements ``peerdns`` and ``dns{1,2}`` are optional for
+all but one NIC.  These are what set your nameservers in
+``/etc/resolv.conf``.
+``dns1`` and ``dns2`` should be defined under a single NIC and ``peerdns``
+should be set to ``"yes"``.  Defined in the
+secrets repo::
+
+    # Example:
+    gw_networks:
+      private:
+        ifname: "eth0"
+        mac: "de:ad:be:ef:12:34"
+        ip4: "192.168.1.100"
+        netmask: "255.255.240.0"
+        gw4: "192.168.1.1"
+        defroute: "yes"
+        peerdns: "yes"
+        search "private.example.com"
+        dns1: "192.168.1.1"
+        dns2: "8.8.8.8"
+      public:
+        ifname: "eth1"
+        etc...
+
+The *fail2ban* vars are explained in /etc/fail2ban/jail.conf.  We've set
+defaults in ``roles/gateway/defaults/main.yml`` but they can be overridden in
+the secrets repo::
+
+    gw_f2b_ignoreip: "127.0.0.1/8"
+    gw_f2b_bantime: "43200"
+    gw_f2b_findtime: "600"
+    gw_f2b_maxretry: "5"
+
+``gw_f2b_services`` is a dictionary listing services fail2ban should monitor.  Defined in
+``roles/gateway/defaults/main.yml``.  See example below::
+
+    gw_f2b_services:
+      sshd:
+        enabled: "true"
+        port: "ssh"
+        logpath: "%(sshd_log)s"
+      apache:
+        enabled: "true"
+        port: "http"
+
 Tags
 ++++
 
@@ -66,6 +127,17 @@ packages
 
 users
     Update OpenVPN users list
+
+networking
+    Configure basic networking (NICs, IP forwarding, resolv.conf)
+
+firewall
+    Configure firewalld
+
+**NOTE:** Ansible v2.1 or later is required for the initial firewall setup as the ``masquerade`` parameter is new to that version.
+
+fail2ban
+    Configure fail2ban
 
 Dependencies
 ++++++++++++
@@ -80,9 +152,6 @@ To Do
 
 - Support installation of new OpenVPN gateway from scratch
 - Generate and pull (to secrets?) CA, keys, and certificates
-- Configure networking
-- Configure firewall
-- Configure fail2ban
 - Configure log rotation
 
 .. _Sepia: https://ceph.github.io/sepia/
--- a/roles/gateway/tasks/fail2ban.yml
+++ b/roles/gateway/tasks/fail2ban.yml
@@ -1,0 +1,41 @@
+---
+- name: Write fail2ban defaults conf file
+  template:
+    src: templates/f2b.jail.local.j2
+    dest: /etc/fail2ban/jail.local
+  notify: restart fail2ban
+
+# Set a var equal to our ansible_managed var since ansible_managed
+# can't be called directly in the next task.
+# See https://github.com/ansible/ansible/issues/11317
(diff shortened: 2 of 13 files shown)


dmick:
We discovered we'd need to update the firewall stuff for port 80/443

djgalloway:
@dmick  Updated to optionally allow http and https

roles/gateway/tasks/fail2ban.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: Write fail2ban defaults conf file
+  template:
+    src: templates/f2b.jail.local.j2
+    dest: /etc/fail2ban/jail.local
+  notify: restart fail2ban
+
+# In case malformed service file was previously shipped.
+# Malformed service files cause fail2ban to not start.
+- name: Clean up ansible-written service conf files
+  shell: for file in $(grep -l "ansible" /etc/fail2ban/jail.d/*); do rm -vf $file; done
```
dmick: just having "ansible" somewhere in the file seems a little dodgy.  Can we grep for {{ansible_managed}"} instead?

roles/gateway/README.rst
```diff
@@ -58,6 +61,63 @@ A list of users that don't have their ssh pubkey added to the ``teuthology_user`
...
+    gw_resolv_search: []
+    # Example: gw_resolv_search: "front.example.com"
+
+    gw_resolv_ns: []
+    # Example:
+    gw_resolv_ns:
+      - 1.2.3.4
+      - 8.8.8.8
+
+The ``gw_networks`` var assumes you have individual "physical" NICs for each
```
dmick: physical doesn't need quotation marks; that implies that it's not really physical in some way.  If you want to emphasize that they must be physical nics, maybe *?

roles/gateway/README.rst
```diff
@@ -58,6 +61,63 @@ A list of users that don't have their ssh pubkey added to the ``teuthology_user`
...
+VLAN in your lab.  The subelements ``peerdns`` and ``dns{1,2}`` are optional.
+``dns1`` and ``dns2`` should be defined under a single NIC and ``peerdns``
+should be set to ``"yes"``.  Defined in the
```
dmick: What do peerdns and dns{1,2} mean?  and if peerdns must be yes, why is it an option?

roles/gateway/README.rst
```diff
@@ -58,6 +61,63 @@ A list of users that don't have their ssh pubkey added to the ``teuthology_user`
...
+The *fail2ban* vars are explained in /etc/fail2ban/jail.conf.  We've set
+defaults in ``roles/gateway/defaults/main.yml`` but they can be overridden in
+the secrets repo::
+
+    gw_f2b_ignoreip: "127.0.0.1/8"
+    gw_f2b_bantime: "43200"
+    gw_f2b_findtime: "600"
+    gw_f2b_maxretry: "5"
+
+A dictionary listing services fail2ban should monitor.  Defined in
```
dmick: Is this missing a variable name?

roles/gateway/README.rst
```diff
@@ -67,6 +127,17 @@ packages
 users
     Update OpenVPN users list
 
+networking
+    Configure basic networking (NICs, IP forwarding, resolv.conf)
+
+firewall
+    Configure firewalld
+
+**NOTE:** Ansible v2.1 or great is required for the initial firewall setup as the ``masquerade`` parameter is new to that version.
```
dmick: s/great/later/

roles/gateway/tasks/main.yml
```diff
@@ -49,3 +49,8 @@
     name: "openvpn@{{ openvpn_server_name }}"
     state: started
     enabled: yes
+
+- name: Write logrotate conf file
+  copy:
+    src: files/openvpn.logrotate
+    dest: /etc/logrotate.d/openvpn
```
dmick: does the service need restarting once you write a new config?

roles/gateway/tasks/firewall.yml
```diff
@@ -0,0 +1,60 @@
+---
+- name: Make sure iptables isn't running
+  service:
+    name: iptables
+    state: stopped
+    enabled: false
+  ignore_errors: true
```
djgalloway: The task returns an error `systemd could not find the requested service "iptables"` when the service isn't installed -- which it shouldn't be.  It's really just an extra step to make sure we're using firewalld and not iptables.

roles/gateway/tasks/fail2ban.yml
djgalloway: Actually, no and I think it's because of https://github.com/ansible/ansible/issues/11317.

I could add a unique identifying string to each fail2ban service file instead..?

roles/gateway/README.rst
djgalloway: I only put quotes around it because in our case, the NICs aren't really physical.  I can remove it altogether.

roles/gateway/README.rst
djgalloway: PEERDNS, in conjunction with DNS{1,2}, adds the nameservers defined by DNS{1,2} to /etc/resolv.conf.  It should only bet set on one NIC.  If set to "yes" on multiple NICs, the last NIC that gets brought up will clobber the first nameservers added.

roles/gateway/README.rst
djgalloway: It's right here: https://github.com/ceph/ceph-cm-ansible/pull/249/files#diff-a52a1791a9e2c8023f940ba001b5ebccR112

roles/gateway/README.rst
djgalloway: Right. Fixed.

roles/gateway/tasks/main.yml
djgalloway: What service?

roles/gateway/README.rst
djgalloway: I think I made it clearer now.

roles/gateway/tasks/firewall.yml
dmick: ok.  sad that the service module can't handle that, but, so be it

roles/gateway/tasks/fail2ban.yml
dmick: why don't we add a role to set a fact to that string and use it, with a big XXX comment mentioning that ansible bug so someone removes it someday

roles/gateway/README.rst
dmick: just so I understand: they're not really physical because you're using a vm, but that illusion is opaque to the code, and this would work on a physical host where they really are physical, right?...

roles/gateway/README.rst
dmick: ah.  If we could add a brief commentary to that effect to inform 1) debian-heads and 2) people who are familiar with sysconfig but forget what those are.  even a little clause like "for update of /etc/resolv.conf" would be enough of a hook

roles/gateway/README.rst
dmick: Still seems better to say "gw_f2b_services: A dictionary..."

roles/gateway/tasks/main.yml
dmick: logrotate

djgalloway: logrotate is called by cron.daily out of the box

roles/gateway/tasks/fail2ban.yml
djgalloway: I tried that too but maybe I'm doing something wrong.  Will keep prodding.

roles/gateway/tasks/main.yml
dmick: OK cool

roles/gateway/README.rst
djgalloway: Well the role only works on CentOS ;) but I did add verbiage to clear things up: https://github.com/ceph/ceph-cm-ansible/pull/249/commits/fb0bbff518677bd4707659ca06d8f31b8891073a#diff-a52a1791a9e2c8023f940ba001b5ebccR79

roles/gateway/README.rst
djgalloway: Right

roles/gateway/README.rst
djgalloway: Fixed

roles/gateway/tasks/fail2ban.yml
djgalloway: Got it: https://github.com/ceph/ceph-cm-ansible/pull/249/commits/6653aa75fb7008c9f236d2593b1d6d1bd1d25ef0#diff-7eca5e065fac0e61340c37a81bf22bdfR11

roles/gateway/tasks/main.yml
dmick: cool

PR #251, opened 2016-06-21
users: Set UID_MIN to 1001

zmc:
When building images with edeploy, UID 1000 was being taken by a
"normal" user; we need to reserve this for the teuthology user.

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -46,6 +46,16 @@
   tags:
     - always
 
+# This is to prevent normal (read: human) users from ending up with UID 1000,
+# which testnodes needs for the teuthology user.
+- name: Set UID_MIN to 1001
+  lineinfile:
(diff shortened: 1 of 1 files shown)


zmc:
This shouldn't affect any previously-deployed hosts.

zmc:
I think I see why this isn't happening on other nodes.

If ``ansible_managed.yml`` is run before ``testnodes.yml``, the ``cm`` user is created before all other users. In sepia at least, ``cm``'s UID is 1100. When users are created without specifying a UID, it appears that the OS chooses a UID equal to "the highest existing UID +1".

I had been just running ``testnodes.yml`` and am now running ``cephlab.yml`` via edeploy, so I'm not hitting this issue anymore. But, anyone who isn't setting ``cm_user_uid`` (which has no default) to something higher than the teuthology user ID (which is hardcoded at 1000), will still hit the issue unless this is applied.

PR #252, opened 2016-06-22
cobbler: testnode post-install rc.local fixes

djgalloway:
- Instead of sleeping for 30 secs and hoping network is up, wait until we can
reliably ping the Cobbler host.
- Give ansible more time to run before wget's built-in 15min timeout is reached.  Otherwise ansible will still run and finish but the lockfile won't get touched.

### DIFF
--- a/roles/cobbler/templates/snippets/cephlab_rc_local
+++ b/roles/cobbler/templates/snippets/cephlab_rc_local
@@ -21,10 +21,15 @@ if [ -e $lockfile ]; then
     exit 0
 fi
 
-# Wait just a bit for the network to come up
-sleep 30
+# Wait until we get 10 ping responses from Cobbler host
+# before calling post-install trigger
+until ping -nq -c10 $http_server
+do
+    echo "Waiting for network"
+    sleep 3
+done
 # Run the post-install trigger a second time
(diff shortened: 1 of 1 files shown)


djgalloway:
@zmc @dmick How do y'all feel about this approach?

dmick:
I very much like the positive lock on "is the network up".  It'd be great if we could come up with something like that for "ansible's done" too

djgalloway:
"Ansible's done" is determined when true is returned after the wget command.  Or do you mean "ansible's done without failures?"

dmick:
is that wget only serving the purpose "block until ansible is finished"?  The plumbing is out of my brain again

zmc:
re: ansible being done, we touch /ceph-qa-ready at the end

dmick:
so where we at here

zmc:
I think this looks good. It's been tested, right?

djgalloway:
I modified rc.local on an already provisioned testnode and ran it, yes.  I suppose I could manually edit that snippet on Cobbler and reprovision.  I'll do that.  DNM yet.

djgalloway:
Tested successfully on smithi004 (CentOS 7.2) and smithi043 (Ubuntu 14.04).

PR #253, opened 2016-06-23
users: Greatly speed up ssh pubkey deployment

zmc:
The speedup depends on changes like https://github.com/ceph/ceph-sepia-secrets/pull/135 - but this PR won't break the current behavior

### DIFF
--- a/roles/users/tasks/update_keys.yml
+++ b/roles/users/tasks/update_keys.yml
@@ -38,9 +38,3 @@
     key: "{{ item.key }}"
   with_items: "{{ pubkey_users }}"
   when: item.key is defined
-  # Register and retry to work around transient githubusercontent.com issues
-  register: ssh_key_update
-  until: ssh_key_update|success
-  # try for 2 minutes to retrieve the key before failing
-  retries: 24
-  delay: 5
--- a/roles/users/tasks/update_keys.yml
+++ b/roles/users/tasks/update_keys.yml
@@ -1,0 +1,12 @@
+---
+- name: Update authorized_keys for each user.
+  authorized_key:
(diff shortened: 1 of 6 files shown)


roles/users/tasks/update_keys.yml
```diff
@@ -1,9 +1,43 @@
...
+
+# The following set_fact calls are apparently necessary to avoid using sudo on
+# localhost to clone the keys repo.
+- set_fact:
+    ansible_become_orig: "{{ ansible_become }}"
+
+- set_fact:
+    ansible_become: false
+
+- name: Clone the keys repo
```
dmick: do we want something to remove the local repo when done?

zmc: I never actually thought of a reason to do that. Can you?

zmc: Actually, I can think of a reason *not* to - another ansible process just cloned and is about to read from it. Removing it would break things

dmick: Well, I was thinking at the end of this ansible-playbook run.   Reason to: meh.  trash left behind, is all.

zmc: Still racy. And, in this case, one job's trash is another's treasure

dmick: also, should we add accept_hostkey because of this:  http://docs.ansible.com/ansible/git_module.html#notes ?

zmc: We're just using https for the repo, so I don't think we need it

dmick: agreed.

PR #266, opened 2016-07-13
use 120s instead of default 12s for cmd timeout

vasukulkarni:
Increase timeout from 12s to 120s for ansible cmds

Signed-off-by: Vasu Kulkarni vasu@redhat.com


ktdreyer:
That's "what" it does, but not "why" :(

djgalloway:
@ktdreyer We were seeing an unusually high number of timeouts in OVH and Octo.

Example: http://sentry.ceph.com/sepia/teuthology/issues/736/events/31647/

ktdreyer:
Thanks for the context!

PR #274, opened 2016-08-01
Use the profile_tasks plugin

zmc:
This will cause ansible to display timing information for tasks,
allowing us to see what's slow and potentially improve speed.

I've been using this locally for a while now. It's nice.


djgalloway:
Is the plugin shipped with ansible already or?  I don't see docs on the official website but found the plugin on github.

I like it.  Just curious if anything else needs to be done.

zmc:
It was in 2.0: https://github.com/ansible/ansible/blob/v2.0.0.0-1/lib/ansible/plugins/callback/profile_tasks.py

djgalloway:
Cool.  Thanks for the info and rebasing.

PR #277, opened 2016-08-05
teuthology: Provide a default for teuthology_yaml_extra

zmc:


djgalloway:
> There appears to be a bug with the free strategy when only using one host.

Can you elaborate?

zmc:
It appeared to get stuck in a loop, not doing anything productive :-/

If you want to see output I can get some

zmc:
I actually can't reproduce now; dropped the commit

PR #278, opened 2016-08-11
Add support for testing RHEL beta distros using internal repos

djgalloway:
Combine with ceph-octo-secrets PR# 77

Tested on RHEL7.2 and RHEL7.3 using stock cobbler images.  RHEL7.2 gets subscribed to CDN as it should and RHEL7.3 uses the internal repos.  Also tested CentOS 7.2 for sanity and to verify `redhat-lsb-core` package is available.

`no_log: true` is ignored (https://github.com/ansible/ansible/issues/11253) for now but since logs are internal, it's not a big deal.  The only 'sensitive' info that gets output is the internal repo path.

### DIFF
--- a/roles/cobbler/templates/snippets/cephlab_packages_rhel
+++ b/roles/cobbler/templates/snippets/cephlab_packages_rhel
@@ -9,6 +9,7 @@
 #end if
 perl
 wget
+redhat-lsb-core
 ## For ansible/NRPE
 smartmontools
 libselinux-python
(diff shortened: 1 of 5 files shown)


djgalloway:
@dmick seemed okay with this.  @zmc any objections?

zmc:
Yeah, looks good! It's been tested?

djgalloway:
See comment 1 :)

PR #279, opened 2016-08-16
defaults: beta_repos must be a dict

jdurgin:
The incorrect default causes ceph-cm-ansible to fail every teuthology job: 'with_dict expects a dict'

Signed-off-by: Josh Durgin <jdurgin@redhat.com>


jdurgin:
merging myself since it's a trivial fix and I tested it with teuthology manually

PR #282, opened 2016-09-13
Add vmhosts to vmlist.py and run virsh in read-only mode

djgalloway:
'virsh -r' allows vmlist.py to work with RHEV hypervisor hosts


dmick:
Look at you!  Cool.

PR #285, opened 2016-10-18
common: Change method of detecting beta distro

djgalloway:
Tested on:
RHEL7.3 Beta
- Unregistered

RHEL7.3 RC2
- Unregistered

RHEL7.2
- Unregistered
- Already registered with expected repos from `rhsm_repos`
- Already registered with additional repos not in `rhsm_repos`

It was recently discovered that Release Candidate ISOs of RHEL don't output "Beta" or
"Alpha" in `lsb_release` output.  Therefore, RHEL RC Distros were
registering with CDN and not putting beta_repos in place.

This new method registers every RHEL system with CDN, checks if there are
any available matching repos based on ansible_distribution_version, and
either:
  - Unregisters from CDN and runs beta_distros.yml if there are no CDN repos
  - Continues with setting up desired rhsm_repos

I also added some additional working to task descriptions to hopefully
claer up what's going on in this playbook since there are a lot of when
and if statements.

### DIFF
--- a/roles/common/tasks/rhel-entitlements.yml
+++ b/roles/common/tasks/rhel-entitlements.yml
@@ -25,7 +25,7 @@
   changed_when: false
   no_log: true
 
-- name: Set rhsm_registered
+- name: Set rhsm_registered if we're already registered
   set_fact:
     rhsm_registered: "{{ subscription.rc == 0 }}"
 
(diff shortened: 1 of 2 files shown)


vasukulkarni:
:+1:

PR #287, opened 2016-11-01
testnode: Define a few empty variables for ansible v2.2 compatibility

djgalloway:
Tested using ansible v2.2 on a smithi running Ubuntu 14.04, a smithi running CentOS 7.2, and a magna running RHEL 7.2.

`roles/testnode/vars/yum_systems.yml` isn't included on playbook runs where `ansible_pkg_mgr != apt`.  Even though the tasks relying on those vars are skipped, the tasks were still fatally failing because the vars weren't defined.  Defining empty defaults gets around that.

### DIFF
--- a/roles/testnode/defaults/main.yml
+++ b/roles/testnode/defaults/main.yml
@@ -29,8 +29,14 @@ epel_packages: []
 # version, distro or package type
 common_packages: []
 
+# common packages that aren't available in aarch64 architecture
+non_aarch64_packages: []
+non_aarch64_packages_to_upgrade: []
+non_aarch64_common_packages: []
+
 # packages used by ceph we want to ensure are removed
 ceph_packages_to_remove: []
+ceph_dependency_packages_to_remove: []
 packages_to_remove: []
 packages_to_upgrade: []
 
--- a/roles/testnode/tasks/apt/repos.yml
+++ b/roles/testnode/tasks/apt/repos.yml
@@ -8,7 +8,7 @@
 
 - name: Remove custom repos
   file: path=/etc/apt/sources.list.d/{{ item }} state=absent
-  with_items: "{{ custom_repos.stdout_lines }}"
+  with_items: "{{ custom_repos.stdout_lines|default([]) }}"
   # Ignore changes here because we will be removing repos that we end up re-adding later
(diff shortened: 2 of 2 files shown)


roles/testnode/defaults/main.yml
```diff
@@ -29,8 +29,14 @@ epel_packages: []
...
 common_packages: []
 
+# common packages that aren't available in aarch64 architecture
+non_aarch64_packages: []
+non_aarch64_packages_to_upgrade: []
+non_aarch64_common_packages: []
+
 # packages used by ceph we want to ensure are removed
 ceph_packages_to_remove: []
+ceph_dependency_packages_to_remove: []
```
zmc: This might be confusing as this is defined in ``roles/testnode/vars/yum_systems.yml``. On the other hand, maybe it's fine. @dmick?

djgalloway: `roles/testnode/vars/yum_systems.yml` isn't included on playbook runs where `ansible_pkg_mgr != apt`.  Even though the tasks relying on those vars are skipped, the tasks were still fatally failing because the vars weren't defined.  Defining empty defaults gets around that.

Not arguing the confusing aspect.  Just explaining why it also needs to be set here.  Otherwise we need to use `|default([])`.

zmc: Yeah, I'm curious what @dmick thinks about how we should set these defaults. I'm not attached to anything in particular.

dmick: if it's yum-specific but in a global place, it'd be better if it had yum in its name IMO

dmick: ...and is one the same as the other, just apt vs yum?  if so I hate the names bad

zmc: I'm confused; these aren't new, they just need defaults to cope with a change in ansible 2.2 regarding undefined variable treatment

dmick: I'm not saying they're new, just, they've now got my attention.  I'm trying to figure out what and why they are, and I'm also confused about why this list wasn't in defaults before (i.e., I suppose, what the convention is for defaults).

PR #288, opened 2016-11-02
common: Install a few additional packages for NRPE

djgalloway:
check_load nagios plugin needs to be installed on RHEL and CentOS


djgalloway:
OK to merge now.  Forgot to add the new var to the README originally.

PR #291, opened 2016-11-16
Updates to downstream_setup role

djgalloway:
Vasu started seeing errors about `repos_to_remove` not being defined after we un-pinned Ansible from 2.1.  This goes back to the issue where task files that aren't even included or run are still being evaluated.  If they don't have empty variables set, the playbook fails.

So I defined an empty list for the var but was still getting:

`FAILED! => {"failed": true, "msg": "{{ repos_to_remove.split(',') }}: 'list object' has no attribute 'split'"}`

I found https://github.com/ansible/ansible/issues/18483 and have set both `setup.yml` and `cleanup.yml` when conditionals to be evaluated dynamically in https://github.com/ceph/ceph-cm-ansible/commit/adbc0903d3944d850be643e1a6dab169090eab7b

Fixes: http://tracker.ceph.com/issues/17922

### DIFF
--- a/roles/downstream-setup/defaults/main.yml
+++ b/roles/downstream-setup/defaults/main.yml
@@ -34,3 +34,6 @@ disable_yum_repos: []
 # NOTE: this does not work on repo files with multiple entries in them,
 # it will only enable the first entry in the repo file.
 enable_yum_repos: []
+
+# defining empty var for ansible v2.2 compatibility.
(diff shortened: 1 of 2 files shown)


vasukulkarni:
http://pulpito.ceph.redhat.com/vasu-2016-11-16_15:13:52-smoke-jewel---basic-clara/

PR #299, opened 2016-12-22
Add role for nsupdate_web

zmc:
See https://github.com/zmc/nsupdate-web/

### DIFF
--- a/roles/nsupdate_web/tasks/main.yml
+++ b/roles/nsupdate_web/tasks/main.yml
@@ -1,0 +1,87 @@
+---
+- name: Build args to pass to nsupdate_web
+  set_fact:
+    nsupdate_web_args: "--ttl {{ nsupdate_web_ttl }} -d {{ lab_domain }} -K {{ keys_dir }}/{{ pubkey_name }} -s {{ nsupdate_web_server }}{% if allow_hosts %} -a {{ allow_hosts }}{% endif %}"
+
+- name: Including major version specific variables.
+  include_vars: "{{ item }}"
+  with_first_found:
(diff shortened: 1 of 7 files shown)


zmc:
Next step is probably an nginx proxy to front the server so it can listen on the main interface - but I need to make changes to `nsupdate-web` to get it to enforce a policy first.

Oh and a README for the role.

djgalloway:
You could probably get it to pull the key from `ansible/secrets/nameserver.yml` if you wanted to.

zmc:
@djgalloway I thought about that but I'm leaning towards that being a not-so-great idea since it'd allow anyone with the key to modify records on any of the dynamic zones. The purpose of the web service is to keep everything locked down

roles/nsupdate_web/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+  user:
+    name: "{{ nsupdate_web_user }}"
+    state: present
+    system: true
+    shell: "/bin/false"
+
+- name: Clone nsupdate_web repo
+  git:
+    repo: "{{ repo_url }}"
+    dest: "~/nsupdate_web"
```
djgalloway: Will the git module create this dir if it doesn't exist?  Ansible docs don't say.

zmc: Yes, it will

zmc:
@djgalloway  This now fronts with nginx. Do you think a readme is necessary?

djgalloway:
I'd say yes.  Mainly to mention that this is a first-time setup role and not really an ongoing config management role.

Also what to do after the role has been run (like putting the keys in place) would be helpful.

zmc:
@djgalloway repushed with a readme and the missing nginx file!

PR #301, opened 2017-01-12
Ensure search domain and domain are set properly in resolv.conf

zmc:


djgalloway:
Does this survive a reboot?  (Does it need to?)

I know @dmick has done some extensive research into how Ubuntu deals with the resolv.conf file.  As I understand things, it's not just a static file you can edit and have the changes be live immediately anymore but I'm not sure.  https://www.stgraber.org/2012/02/24/dns-in-ubuntu-12-04/

zmc:
@djgalloway Yeah, it should.

This is why we [purge resolvconf](https://github.com/ceph/ceph-cm-ansible/blob/a6f7b4d1c4685c647216d03e7d6bb82ce8a71172/roles/testnode/tasks/resolvconf.yml#L2-L7)

PR #304, opened 2017-02-04
Fix some bugs uncovered during testing

smithfarm:


smithfarm:
@zmc @dmick

roles/testnode/tasks/apt_systems.yml
```diff
@@ -3,7 +3,7 @@
 # For anything ubuntu < 15.04 or debian <=7 we still do.
 - name: Setup local repo files.
   include: apt/repos.yml
-  when: ansible_distribution_major_version|int < 15
+  when: ansible_pkg_mgr == "apt" and ansible_distribution_major_version|int < 15
```
dmick: I would have assumed we're only doing apt_systems.yml when pkg_mgr == apt; no?

smithfarm: I would have, too, but I found that it gets run when ansible_pkg_mgr == zypper.

dmick: I think we need to know why that is.  The separation is sort of fundamental.

dmick: http://docs.ansible.com/ansible/playbooks_conditionals.html#applying-when-to-roles-and-includes seems to imply that it might evaluate them all, but they should all get the same 'when' conditional, no?

smithfarm: @dmick So, add a `when:` with the same conditional to all the blocks in the file where it is currently missing?

dmick: the way I read that, ansible is supposed to apply the when: from the include to every task in the included file.  I haven't constructed a test.

PR #305, opened 2017-02-06
testnode/tasks/ntp: remove '/etc/network/if-up.d/ntpdate' on Xenial

dmick:
It's buggy (see comment in code) and we don't need it, because
teuthology does what it would do anyway.  Supposed to be fixed in
a later version of the Debian package which is in Yakkety.

Fixes: http://tracker.ceph.com/issues/18837


dmick:
Prerequisite for https://github.com/ceph/teuthology/pull/1013

PR #307, opened 2017-02-23
public_facing role

djgalloway:
This role will probably be forever-growing but it's at a point, now, where fail2ban and UFW are at least managed on all hosts.

### DIFF
--- a/roles/public_facing/tasks/download.ceph.com.yml
+++ b/roles/public_facing/tasks/download.ceph.com.yml
@@ -1,0 +1,69 @@
+---
+- name: Put logrotate config in place
+  template:
+    src: templates/download.ceph.com/logrotate.j2
+    dest: /etc/logrotate.d/download.ceph.com
+
+# Used for pushing upstream builds
+# https://github.com/ceph/ceph-build/blob/master/scripts/sync-push
+- name: Add signer user
+  user:
+    name: signer
+
+# signer_pubkey defined in inventory host_vars
+- name: Update signer user's authorized_keys
+  authorized_key:
+    user: signer
+    state: present
+    key: "{{ signer_pubkey }}"
+
+# Used to rsync download.ceph.com http logs and compile metrics
+# for metrics.ceph.com
+- name: Create Bitergia user
+  user:
+    name: bitergia
+    groups: adm
+
+# bitergia_pubkey defined in inventory host_vars
+- name: Update bitergia user's authorized_keys
+  authorized_key:
+    user: bitergia
+    state: present
+    key: "{{ bitergia_pubkey }}"
+
+- name: Create ~/bin dir for bitergia user
+  file:
+    path: /home/bitergia/bin
+    state: directory
+    owner: bitergia
+    group: bitergia
+
+# Rsync is almost certainly already installed but it's required for the next task
+- name: Make sure rsync is installed
+  apt:
+    name: rsync
+    state: latest
+
+- name: Put rrsync script in place for bitergia user
+  shell: "gunzip /usr/share/doc/rsync/scripts/rrsync.gz --to-stdout > /home/bitergia/bin/rrsync"
+  changed_when: false
+
+- name: Set permissions for bitergia rrsync script
+  file:
+    dest: /home/bitergia/bin/rrsync
+    owner: bitergia
+    group: bitergia
+    mode: 0774
+
+# Updates download.ceph.com/timestamp
+- name: Put make_timestamp script in place
+  template:
+    src: templates/download.ceph.com/make_timestamp.j2
+    dest: /usr/libexec/make_timestamp
+    mode: 0775
+
+- name: Create cron entry for make_timestamp
+  cron:
+    name: "Update download.ceph.com/timestamp"
+    minute: "0"
+    job: "/usr/libexec/make_timestamp"
--- a/roles/public_facing/tasks/fail2ban.yml
+++ b/roles/public_facing/tasks/fail2ban.yml
@@ -1,0 +1,77 @@
+---
+- name: Install or update fail2ban
+  package:
+    name: fail2ban
+    state: latest
+
+- name: Check if firewalld is running
+  shell: firewall-cmd --state
+  register: firewalld_status
+  # Don't fail if command not found
+  failed_when: false
+
+- name: Set f2b_banaction if using firewalld
+  set_fact:
+    f2b_banaction: "firewallcmd-ipset"
+  when: firewalld_status.stdout == "running"
+
+- name: Check if UFW is running
+  shell: ufw status | grep Status | cut -d ' ' -f2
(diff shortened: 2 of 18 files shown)


roles/public_facing/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+# ansible_host changes to 'localhost' in the next task.  Set target_host
```
dmick: wat

djgalloway: When you run something with `local_action:`, the `ansible_host` var temporarily changes to 'localhost'

dmick: oh, because local_action means 'it's running on localhost', but you need to have the name of the ultimate target host? If so, just change this comment to mention "local_action in the next task makes ansible_host be the local hostname" (or 'localhost' if that was a literal statement)

roles/public_facing/files/download.ceph.com/rrsync
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/perl
```
dmick: It still bugs me that we're including a custom one rather than the one that comes with rsync.  If we're really gonna, we ought to insert a comment saying that we are, and why, so that someone looking doesn't get confused in a year

roles/public_facing/tasks/download.ceph.com.yml
```diff
@@ -0,0 +1,52 @@
...
+    src: download.ceph.com/rrsync
+    dest: /home/bitergia/bin/rrsync
+    owner: bitergia
+    mode: 0774
+
+# Updates download.ceph.com/timestamp
+- name: Put make_timestamp script in place
+  copy:
+    src: download.ceph.com/make_timestamp
+    dest: /var/tmp/make_timestamp
```
dmick: I don't care that it's in /var/tmp, btw, that's just my manual 'goto' for things that don't go away on boot.  /usr/libexec might make more sense, or something.

dmick:
LGTM otherwise.

roles/public_facing/files/download.ceph.com/rrsync
djgalloway: I'm waiting to hear back from the bitergia folks if we can just use the version that comes with the rsync package.

roles/public_facing/tasks/download.ceph.com.yml
djgalloway: Noted.  Will move to somewhere more appropriate... `/usr/local/bin/` perhaps?

dmick: I've never really understood why the smart commands etc go in /usr/libexec, but I'd prefer there unless there's a good reason just because Precedent

mscherer: /var/tmp/ is also writable by all users, so that's a potential security issue (albeit unlikely to be one in practice). And for /usr/libexec, that's some old historic deviation from FHS from RH.

roles/public_facing/tasks/fail2ban.yml
```diff
@@ -0,0 +1,77 @@
...
+  notify: restart fail2ban
+
+# sshd_logpath is used in the f2b_services dictionary.  fail2ban doesn't know
+# where ssh logs are for services other than sshd so sshd-ddos, for example
+# needs to be told where to look.  For other services (e.g., nginx), the logpath
+# can be set directly in the dict.
+- name: Set sshd_logpath for CentOS/RHEL
+  set_fact:
+    sshd_logpath: "/var/log/messages"
+  when: ansible_pkg_mgr == "yum"
```
mscherer: I would suggest to use ansible_os_family, since this would avoid issue if the package manager switch to DNF (ie, Fedora, or a newer version of RHEL/Centos)

roles/public_facing/tasks/fail2ban.yml
```diff
@@ -0,0 +1,77 @@
...
+
+- name: Set sshd_logpath for Ubuntu
+  set_fact:
+    sshd_logpath: "/var/log/auth.log"
+  when: ansible_pkg_mgr == "apt"
+
+# This makes sure there are no old or malformed service conf files.
+# We'll rewrite them in the next task.
+- name: Clean up local service conf files
+  shell: rm -f /etc/fail2ban/jail.d/*.local
```
mscherer: Shouldn't you also reload fail2ban here ? If a file is removed, but nothing new is added in /etc, this will not trigger a configuration change.

roles/public_facing/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+  tags:
+    - always
+  when: use_fail2ban == true
+
+- name: Disable password authentication
+  lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: "^PasswordAuthentication"
+    line: "PasswordAuthentication no"
+    state: present
```
mscherer: You may need to restart/reload openssh here, or it would let server in insecure state.

roles/public_facing/templates/status.sepia.ceph.com/nagios-eventhandler-cachet.config.j2
```diff
@@ -0,0 +1,6 @@
+<?php
+
+$cachet_url = '{{ cachet_api_url }}';
+$api_key = '{{ cachet_api_key }}'
+
```
mscherer: This would later leak in diff, it may be something to watch for and add a no_log:True (in case you plan later to have a public instance of ara or jenkins to deploy, this kind of things)

roles/public_facing/tasks/fail2ban.yml
djgalloway: No because `f2b_services` is a required dict so the next task will always write new conf files.  It's sort of an overkill method of making sure the *only* conf files that exist are the ones that get written in the next task.

roles/public_facing/tasks/main.yml
djgalloway: Fair point.  Thanks.

roles/public_facing/templates/status.sepia.ceph.com/nagios-eventhandler-cachet.config.j2
djgalloway: Not a huge deal.  This role will only be run by humans.

roles/public_facing/files/download.ceph.com/rrsync
djgalloway: They unexplicitly said it didn't matter.

djgalloway:
@dmick I think everything's been addressed.  I moved everything that was in `roles/public_facing/files` to the `templates` dir so `{{ ansible_managed }}` would propagate.  I missed that before.

PR #308, opened 2017-02-27
Reduce 'internal_poll_interval' to stem the CPU usage tide

dmick:
See https://github.com/ansible/ansible/pull/16560 and
https://github.com/ansible/ansible/issues/14219


dmick:
I had this in place by hand for a few days; I wasn't very analytical about measuring anything but I didn't really notice any difference in runs except that they weren't constantly at the top of the htop list (and I saw them calling select with 10000 µs instead of 1000)

PR #310, opened 2017-03-07
common: Reinstall selinux-policy-targeted if needed

djgalloway:
This ensures module file integrity so we can build the nrpe selinux
module in the proceeding tasks.

Fixes: http://tracker.ceph.com/issues/19126


djgalloway:
It's not an ideal solution since the root cause is still unknown but I've confirmed it gets a machine that's in a bad state back to a clean state.

dmick:
Why not?

PR #318, opened 2017-05-24
Support NVMe drive monitoring via Nagios

djgalloway:
NVMe support works.  I was holding off on filing a PR until I added logic for SATA SSDs but that's going to have to wait for now.

http://tracker.ceph.com/issues/19346


roles/common/files/libexec/smart.sh
```diff
@@ -103,6 +110,17 @@ preflight ()
...
   fi
+
+  # Check for nvme devices and nvme-cli executable
+  if cat /proc/partitions | grep -q nvme
+  then
+    nvme=true
+    if ! [ -x "$nvmecli" ]
+    then
+      echo "ERROR - NVMe Device detected but no nvme-cli executable"
+      exit 3
```
dmick: exit here, or continue trying to check spinners/ssds?

roles/common/files/libexec/smart.sh
```diff
@@ -288,5 +306,72 @@ normal_smart ()
...
+nvme_smart ()
+{
+  # Loop through NVMe devices
+  for nvmedisk in $(sudo $nvmecli list | grep nvme | awk '{ print $1 }')
+  do
+    # Include NVMe devices in overall drive count
+    let "numdrives+=1"
+    # Clear output variable from any previous disk checks
+    output=""
+    output=$(sudo $nvmecli smart-log $nvmedisk | \
```
dmick: man, this is painful.  No more friendly output format?

dmick: also, did you consider smart-log-add and reject it?  I don't see that there's anything more useful there, but, just wondering if you noticed

dmick:
Looking reasonably good.  Just a few discussion comments.

roles/common/files/libexec/smart.sh
djgalloway: Hm good question.  I think the reasoning I had for it exiting was to make sure it alerted, and thus got looked at and fixed.

Otherwise, if spinners get checked and are healthy, it'll stay green in nagios but won't indicate any problem with the executable being missing.

roles/common/files/libexec/smart.sh
djgalloway: Yeah it's basically doing the same thing we're doing with grepping `cli64 disk smart drv=X` output.

I don't think I'd actually looked at smart-log-add.  Here's output comparison.

```
root@smithi163:~# nvme intel smart-log-add /dev/nvme0n1
Additional Smart Log for NVME device:nvme0n1 namespace-id:ffffffff
key                               normalized raw
program_fail_count              : 100%       0
erase_fail_count                : 100%       0
wear_leveling                   : 100%       min: 140, max: 145, avg: 142
end_to_end_error_detection_count: 100%       0
crc_error_count                 : 100%       0
timed_workload_media_wear       : 100%       63.999%
timed_workload_host_reads       : 100%       65535%
timed_workload_timer            : 100%       65535 min
thermal_throttle_status         : 100%       0%, cnt: 0
retry_buffer_overflow_count     : 100%       0
pll_lock_loss_count             : 100%       0
nand_bytes_written              : 100%       sectors: 2426197
host_bytes_written              : 100%       sectors: 2143516

root@smithi163:~# nvme smart-log /dev/nvme0n1
Smart Log for NVME device:nvme0n1 namespace-id:ffffffff
critical_warning                    : 0
temperature                         : 25 C
available_spare                     : 100%
available_spare_threshold           : 10%
percentage_used                     : 0%
data_units_read                     : 23,528,053
data_units_written                  : 140,477,521
host_read_commands                  : 237,408,541
host_write_commands                 : 1,566,107,255
controller_busy_time                : 43
power_cycles                        : 22
power_on_hours                      : 3,176
unsafe_shutdowns                    : 20
media_errors                        : 0
num_err_log_entries                 : 0
Warning Temperature Time            : 0
Critical Composite Temperature Time : 0
```

I don't think I'd use it in place of but it could be useful in addition to `smart-log`.  Hard to tell without seeing output of an actual failing drive.  What do you think?

roles/common/files/libexec/smart.sh
```diff
@@ -143,7 +143,8 @@ areca_smart ()
         messages+=("Drive $slot $dl has $reallocated reallocated sectors")
         failed=true
         # A small number of reallocated sectors is OK
-        if [ "$reallocated" -le 5 ]
+	# Don't set rc to WARN if we were already CRIT from previous drive
+        if [ "$reallocated" -le 5 ] && [ "$rc" != 2 ]
```
dmick: logic starts being icky.  If we have to touch this again, might consider something like calling a "increase()" function to only increase rc, so you don't need if[] everywhere, or early-out from the tests if you have any fails, or something

roles/common/files/libexec/smart.sh
dmick: probably ok.

roles/common/files/libexec/smart.sh
dmick: shrug.  just that it has extra info is all I think I meant (who knows, it was months ago)

PR #326, opened 2017-07-31
testnodes: Clean up leftover Ceph artifacts from previous jobs

djgalloway:
Fixes: http://tracker.ceph.com/issues/20623

### DIFF
--- a/roles/testnode/tasks/main.yml
+++ b/roles/testnode/tasks/main.yml
@@ -97,6 +97,24 @@
   tags:
     - resolvconf
 
+# http://tracker.ceph.com/issues/20623
+- name: List any leftover Ceph artifacts from previous jobs
+  shell: 'find {{ item }} -name "*ceph*"'
+  with_items:
+    - /var/run/
+    - /etc/systemd/system/
+    - /etc/ceph
+    - /var/log/
+  register: ceph_test_artifacts
+  changed_when: ceph_test_artifacts.stdout != ""
+  failed_when: ceph_test_artifacts.rc != 0 and "No such file or directory" not in ceph_test_artifacts.stderr
+
+- name: Delete any leftover Ceph artifacts from previous jobs
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items: "{{ ceph_test_artifacts.results|map(attribute='stdout_lines')|list }}"
+
 # Touch a file to indicate we are done. This is something chef did;
 # teuthology.task.internal.vm_setup() expects it.
 - name: Touch /ceph-qa-ready


djgalloway:
Marked DNM because I'm just interested in review/opinion on method used here for now.

vasukulkarni:
👍  very much needed! , Not sure with_items will remove the empty dir's as well in the end.

djgalloway:
@vasukulkarni Yeah, the other option would be to just use the shell module to recursively delete paths/dirs.  For example

```
shell: 'rm -rf {{ item }}'
with_items:
  - "/etc/ceph"
  - "/etc/systemd/system/multi-user.target.wants/ceph*"
  - "/var/run/ceph*"
```

But using `find` to recursively find files and dirs to delete seemed more elegant IMO especially when it comes to traversing `/etc/systemd/system`.

djgalloway:
I installed then apt removed ceph on a Xenial machine and this is all the stuff left behind just in `/etc`

```
# find /etc/ -name "*ceph*"
/etc/default/ceph
/etc/logrotate.d/ceph-common
/etc/rc0.d/K02ceph
/etc/pm/sleep.d/25-ceph
/etc/init/ceph-rbd-mirror-all.conf
/etc/init/ceph-mon-all.conf
/etc/init/ceph-osd.conf
/etc/init/ceph-rbd-mirror.conf
/etc/init/ceph-mon-all-starter.conf
/etc/init/ceph-all.conf
/etc/init/ceph-rbd-mirror-all-starter.conf
/etc/init/ceph-osd-all.conf
/etc/init/ceph-osd-all-starter.conf
/etc/init/ceph-disk.conf
/etc/init/ceph-create-keys.conf
/etc/init/ceph-mon.conf
/etc/rc2.d/S04ceph
/etc/rc3.d/S04ceph
/etc/rc6.d/K02ceph
/etc/rc4.d/S04ceph
/etc/bash_completion.d/ceph
/etc/systemd/system/ceph-osd.target
/etc/systemd/system/ceph-mon.target
/etc/systemd/system/multi-user.target.wants/ceph-osd.target
/etc/systemd/system/multi-user.target.wants/ceph-mon.target
/etc/systemd/system/multi-user.target.wants/ceph.target
/etc/systemd/system/ceph.target.wants
/etc/systemd/system/ceph.target.wants/ceph-osd.target
/etc/systemd/system/ceph.target.wants/ceph-mon.target
/etc/rc5.d/S04ceph
/etc/ceph
/etc/rc1.d/K02ceph
/etc/init.d/ceph
```

vasukulkarni:
It wasn't that horrible on centos, :( ,  +1 for find , also looking at the above output,  I believe it should be /etc now instead of /etc/ceph or /etc/systemd?

vasukulkarni:
Do you want to remove the  DNM here ?

djgalloway:
Still curious what @zmc and @dmick think of the method.  e.g., is it safe / is there a saner way?

dmick:
I don't understand why the systemd and rc.d stuff isn't removed on package removal; that seems like a packaging bug we should fix.  AFAIK those are not 'configuration files' that would be held to remove on a 'package purge'.

But even if they were, are we not doing a 'purge' when we remove (on Debian?  I don't know what the equivalent yum/dnf mech is exactly).

in fact...everything here except /etc/ceph should be removed by the package remove.  What's going wrong?

djgalloway:
@dmick : @zmc reminded me `apt` has `purge` and `remove`.  I did a `remove` in the test above which would explain why so much got leftover.

yum is still going to leave stuff behind because there is no `apt-get purge` equivalent.

I re-tested on Xenial using `purge` and way less gets left behind.

```
root@vpm005:~# history
    1  apt install ceph-osd
    2  apt install ceph
    3  apt-get purge ceph
    4  updatedb
    5  locate ceph

<snipped to include relevant paths we want to remove with this PR>
/etc/default/ceph
/etc/logrotate.d/ceph-common
/etc/systemd/system/multi-user.target.wants/ceph.target
/lib/systemd/system/ceph.target
```

Yuri also ran into a few testnodes over the weekend where `/var/log/` didn't get cleaned up and the root disk was almost full.  This PR would be a good candidate to take care of that as well.

This is all until we can move to reimaging testnodes on every job of course.

PR #328, opened 2017-08-14
public_facing: Delete rsync dotfiles on d.c.c

djgalloway:
rsync dotfiles are causing rsync errors when our d.c.c mirrors are
rsyncing new packages

### DIFF
--- a/roles/public_facing/tasks/download.ceph.com.yml
+++ b/roles/public_facing/tasks/download.ceph.com.yml
@@ -68,4 +68,15 @@
     minute: "0"
     job: "/usr/libexec/make_timestamp"
 
+# Occasionally rsync dot file get leftover either on the signer box or on d.c.c.
+# The dotfiles cause issues for d.c.c mirrors because the script just quits instead of skipping files it can't access.
+# https://www.spinics.net/lists/ceph-users/msg37976.html
+#
+# The -mmin +120 is so the cronjob doesn't delete files that are in the process of being rsynced (like during a release).
+- name: Cron entry for deleting rsync dotfiles
+  cron:
+    name: "Delete rsync dotfiles"
+    hour: "0"
+    job: 'find /data/download.ceph.com/www -type f -name ".[^.]*" -mmin +120 -delete'
+
 - import_tasks: letsencrypt_nginx.yml


dmick:
So, what worries me about this is the mostly the incapability of distinguishing what might be "normal" dotfiles (let's say we had .htaccess files or something) from rsync trash.

I note that rsync supports -T for putting tmp dotfiles in a directory.  I wonder who all is making rsync requests of d.c.c and if they could be convinced to use -T?

Another sort of icky thing is "if it's a 50GB upload that fails at 49GB, you really *do* want the upload to resume rather than restart".  I understand it's only a midnight cron, but still, that's exactly the sort of thing that would annoy you at midnight (UTC, I assume), and/or it's entirely possible that saving that time the next day is still useful.

how widespread was this?  what was causing the interruptions of the transfers?  Do we have a feel for how big a problem it is, is what I'm driving at...

And what did it break?  Just running out of space on the partition?

djgalloway:
> So, what worries me about this is the mostly the incapability of distinguishing what might be "normal" dotfiles (let's say we had .htaccess files or something) from rsync trash.

I hear you.  I've seen this a few times before now and the only dotfiles that are ever in that directory are rsync leftovers.  I know this was just an example but we use nginx so we wouldn't have .htaccess files there.

> I note that rsync supports -T for putting tmp dotfiles in a directory. I wonder who all is making rsync requests of d.c.c and if they could be convinced to use -T?

I don't think those dotfiles matter to people rsyncing **from** d.c.c.  I'm pretty sure they just let an rsync process **to** d.c.c pick up where it left off.  So we could just update https://github.com/ceph/ceph-build/blob/master/scripts/sync-push to use -T.  That is an option.

> how widespread was this? what was causing the interruptions of the transfers? Do we have a feel for how big a problem it is, is what I'm driving at...
> And what did it break? Just running out of space on the partition?

Our d.c.c mirror maintainers were seeing errors when running https://github.com/ceph/ceph/blob/master/mirroring/mirror-ceph.sh.  I tested it and instead of skipping over the files rsync didn't have permission to touch, the rsync process would quit.

I can only assume the dotfiles are getting leftover either from broken rsync processes from chacra or to download.ceph.com.  I'd lean toward "someone intentionally killed the process" vs. network disruptions.

I cleaned up all the offending dotfiles on the 7th (https://www.spinics.net/lists/ceph-users/msg37976.html) and got an e-mail on Monday saying another had popped up.

I went with this method because the other options I saw were:

1. Expect release maintainers to not kill rsync processes and to clean up after themselves if they do (but humans)
2. chown that whole dir in cron or after every rsync **to** d.c.c (potential security issue)

dmick:
This doesn't seem to be happening anymore; closing, can reopen if required

PR #329, opened 2017-08-22
testnode: edit lvm.conf on both ubuntu and centos

idryomov:


roles/testnode/tasks/setup-ubuntu.yml
```diff
@@ -12,6 +12,6 @@
   lineinfile:
     dest: /etc/lvm/lvm.conf
     regexp: "# types ="
-    line: 'types = [ "rbd", 1024 ]'
+    line: 'types = [ "rbd", 16 ]'
```
djgalloway: Originally set here for posterity: https://github.com/ceph/ceph-qa-chef/pull/12/files

djgalloway:
lvm2 should get added here https://github.com/ceph/ceph-cm-ansible/blob/master/roles/testnode/vars/centos_7.yml#L18 and here https://github.com/ceph/ceph-cm-ansible/blob/master/roles/testnode/vars/redhat_7.yml#L18

roles/testnode/tasks/setup-ubuntu.yml
idryomov: For xfstests purposes, the value isn't actually important, so I'm not worried about some nodes having 1024 and other nodes having 16.

djgalloway:
Other than adding lvm2 to the package lists, looks good.

```
dgalloway@w541 ceph-cm-ansible (wip-krbd-lvm-conf)$ ansible-playbook testnodes.yml --limit="smithi034*" --tags="lvm" --skip-tags="always" --check --diff

PLAY [testnodes] ****************************************************************************************************************************************************************************************************************************
Wednesday 23 August 2017  12:32:36 -0400 (0:00:00.259)       0:00:00.259 ****** 
--- before: /etc/lvm/lvm.conf (content)
+++ after: /etc/lvm/lvm.conf (content)
@@ -174,7 +174,7 @@
 	# maximum number of partitions.
 	# 
 	# Example
-	# types = [ "fd", 16 ]
+types = [ "rbd", 16 ]
 	# 
 	# This configuration option is advanced.
 	# This configuration option does not have a default value defined.


TASK [testnode : Edit lvm.conf to support LVM on kRBD.] *************************************************************************************************************************************************************************************
changed: [smithi034.front.sepia.ceph.com]

PLAY RECAP **********************************************************************************************************************************************************************************************************************************
smithi034.front.sepia.ceph.com : ok=1    changed=1    unreachable=0    failed=0   

Wednesday 23 August 2017  12:32:38 -0400 (0:00:01.872)       0:00:02.131 ****** 
=============================================================================== 
testnode : Edit lvm.conf to support LVM on kRBD. ------------------------ 1.87s
```

idryomov:
Isn't lvm2 part of Base group?

djgalloway:
> Isn't lvm2 part of Base group?

So it is.  Fair enough.

PR #331, opened 2017-09-01
tools: Script to simplify Cobbler user creation

djgalloway:

### DIFF
--- a/tools/cobbler-access.sh
+++ b/tools/cobbler-access.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+# Script to generate Cobbler credentials
+
+tmpfile=$(mktemp)
+
+# Basically `mkpasswd` but uses a small subset of special characters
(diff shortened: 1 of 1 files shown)


tools/cobbler-access.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# Script to generate Cobbler credentials
+
+tmpfile=$(mktemp)
+
+# Basically `mkpasswd` but uses a small subset of special characters
+password=$(head /dev/urandom | tr -dc 'A-Za-z0-9!@#$%&' | head -c 12 && echo)
```
zmc: What about `pwgen`?

djgalloway: Never heard of it.  Also not pre-installed on Fedora.

Seems silly to install a dependency for a script like this.  Just my opinion though.

zmc: pwgen is like 16kb and is really useful, but this is fine too

PR #332, opened 2017-09-12
Move scripts to help manage items in crontab on the teuthology box

yuriw:
Replaces https://github.com/ceph/teuthology/pull/1108

We have these scripts in `ceph` tree (`ceph/qa/nightlies/` and `ceph/qa/machine_types/`) and used to keep check out directory of `ceph-qa-suites` on the teuthology box in order to use with nightlies.  I think it'd be easier if we move them, and keep pulling latest for teuthology. 

Signed-off-by: Yuri Weinstein <yweinste@redhat.com>

### DIFF
--- a/roles/teuthology/files/vps.yaml
+++ b/roles/teuthology/files/vps.yaml
@@ -1,0 +1,14 @@
+overrides:
+  ceph:
+    conf:
+      global:
+        osd heartbeat grace: 100
+        # this line to address issue #1017 
+        mon lease: 15
+        mon lease ack timeout: 25
+  s3tests:
+    idle_timeout: 1200
+  ceph-fuse:
+    client.0:
+       mount_wait: 60
+       mount_timeout: 120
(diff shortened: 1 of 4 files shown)


yuriw:
@zmc @djgalloway for some reason I can't assign Reviewers, pls tale a look

roles/teuthology/files/vps.yaml
```diff
@@ -0,0 +1,14 @@
+overrides:
```
zmc: Remind me why this file can't be in ceph.git/qa/suites/ ?

yuriw: @zmc because the way we use it I believe per @liewegas ' initial suggestion is we add `vps.yaml` on every line in crontab that runs on `vps` nodes

zmc: That doesn't really answer the question :-/

PR #334, opened 2017-09-15
Make selinux policy module checking bulletproof

djgalloway:


dmick:
Do we have any idea what's leaving zerolength files around?  Is selinux really that dumb?

djgalloway:
@dmick looks ilke we've gotten as far as "weird, can't reproduce" then the bug got EOL'ed.

https://bugzilla.redhat.com/show_bug.cgi?id=784184

Here's another: https://bugzilla.redhat.com/show_bug.cgi?id=1406874

PR #346, opened 2017-11-10
cobbler: Add logic to rc.local for FOG-provisioned machines

djgalloway:
This is for use with FOG.  When we create an OS image, the image
inherits the testnode's hostname that was used to create the image.

This rc.local snippet will make sure the testnode's hostname is set to
whatever the IP's PTR record is on every boot.

`rclocal_nameserver` should be set in the inventory groupvars.

### DIFF
--- a/roles/cobbler/templates/snippets/cephlab_rc_local
+++ b/roles/cobbler/templates/snippets/cephlab_rc_local
@@ -85,8 +85,15 @@ if [ -n "$myips" ]; then
   done
 fi
 
-#end raw
 {% endif %}
+
+# Regenerate SSH host keys on boot if needed
+if command -v apt-get &>/dev/null; then
+  if [ ! -f /etc/ssh/ssh_host_rsa_key ]; then
+     dpkg-reconfigure openssh-server
+  fi
+fi
+#end raw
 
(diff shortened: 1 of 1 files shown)


djgalloway:
https://github.com/ceph/ceph-sepia-secrets/pull/276

roles/cobbler/templates/snippets/cephlab_rc_local
```diff
@@ -9,13 +9,45 @@
 #set script = '/etc/rc.local'
 #end if
 
-cat > $script << EOF
+cat > $script <<\EOF
```
zmc: What's this?

djgalloway: Without the backslash, $(these) are interpreted and run as commands.  https://stackoverflow.com/a/17983608

zmc:
Gonna wait till Monday

djgalloway:
This is good to go.  We also need some logic to fix `/etc/network/interfaces` on Ubuntu and `/etc/sysconfig/network-scripts/ifcfg` scripts in CentOS but I'll do that in a separate PR.

PR #348, opened 2017-11-13
testnode: Avoid an unmountable /var/lib/ceph

zmc:
We've been seeing a lot of this:
http://tracker.ceph.com/issues/21989

I believe the issue is the XFS filesystem being created by tools that default to a version 5 superblock - a version too new for some of our testing kernels.

This is not widely tested - we can talk about it when you're back


djgalloway:
👍

PR #365, opened 2017-12-20
all: Use import_tasks vs include_tasks so tags work again

djgalloway:
See https://github.com/ansible/ansible/pull/33468

http://pulpito.ceph.com/dgalloway-2017-12-20_00:42:54-smoke-luminous-distro-basic-smithi/

### DIFF
--- a/roles/cobbler/tasks/main.yml
+++ b/roles/cobbler/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- include_tasks: ipmi_secrets.yml
+- import_tasks: ipmi_secrets.yml
   tags:
     - always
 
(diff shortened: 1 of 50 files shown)


zmc:
The ansible PR isn't merged, @djgalloway

djgalloway:
@zmc Right, this is fixing our playbooks _because_ the PR isn't merged.

Regardless, we should be using `import` vs `include` anyway since the inclusions aren't dynamic (e.g., `include_tasks: {{ ansible_pkg_mgr }}_systems.yml` vs. what we use: `import_tasks: apt_systems.yml`)

I'm happy to do some cleanup later and switch to using `include` but right now, we can't call a set of tasks using tags because of the ansible PR not being merged.

djgalloway:
I probably should've linked to the issue: https://github.com/ansible/ansible/issues/32015

PR #370, opened 2018-01-17
testnode: Check for and mark down systems with missing NVMe devices

djgalloway:


djgalloway:
http://pulpito.ceph.com/dgalloway-2018-01-17_19:21:56-smoke-luminous-distro-basic-smithi/

PR #373, opened 2018-01-23
ntp-server: Create NTP server role

djgalloway:

### DIFF
--- a/roles/ntp-server/tasks/main.yml
+++ b/roles/ntp-server/tasks/main.yml
@@ -1,0 +1,119 @@
+---
+- name: Check if ntp package installed
+  command: rpm -q ntp
+  ignore_errors: true
+  register: ntp_installed
+
+- name: Check if chrony package installed
+  command: rpm -q chrony
+  ignore_errors: true
+  register: chrony_installed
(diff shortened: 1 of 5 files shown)


roles/ntp-server/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
...
+  when: use_chrony == true
+
+- name: Write NTP config file
+  template:
+    src: ntp.conf.j2
+    dest: /etc/ntp.conf
+  register: conf_written
+  when: use_ntp == true
+
+- name: Write chronyd config file
```
vasukulkarni: Does this write the file for all rpm systems when chronyd is enabled? 

@ktdreyer do you have any comments here? ceph-ansible uses ntp i guess which we can make as standard here as well.

djgalloway: @vasukulkarni this is only for NTP *servers*.   You don't need to worry about testnodes.

vasukulkarni: sorry forgot, it is doing much more than testnodes 👍

PR #214, opened 2016-03-18
A role for PCP

zmc:
This is still a work in progress.

Lots of information on PCP is available here:
http://pcp.io/docs/guide.html

PR #380, opened 2018-02-26
Add support for Red Hat Satellite

djgalloway:


djgalloway:
I just realized adding `common` as a role to `pcp.yml` makes the `common` role run twice which we obviously don't want.

Running the PCP role before the common role on a RHEL Sepia node results in playbook failure because it tries to install PCP when the machine's not subscribed yet.

I thought about moving `pcp.yml` after `testnodes.yml` but it was moved higher for a reason at some point: https://github.com/ceph/ceph-cm-ansible/pull/214/commits/65e2fe0c3b904a648fda1cf98e5647e0799d6653

PR #394, opened 2018-04-09
Always set hardware clock

djgalloway:
@dmick - I know you spent a lot of time looking into NTP and hardware clocks.  Do you recall anything that might make this problematic?

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -25,6 +25,11 @@
 - import_tasks: zypper_systems.yml
   when: ansible_pkg_mgr == "zypper"
 
+- name: Set the hardware clock
+  command: hwclock --systohc
+  tags:
+    - timezone
+
 # configure Kerberos
 - import_tasks: kerberos.yml
   tags:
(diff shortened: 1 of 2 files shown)


djgalloway:
We're using our own internal nameservers now.

dmick:
ok, so: 1) can we possibly get here without a reliably-good system clock? 2) why do we need to set the hwclock (i.e. what's the motivation)?

djgalloway:
This sort of answers both questions.  From the Red Hat docs:

```
 Either you can set the hardware clock to the current system time by using this command:

hwclock --systohc

Note that if you use NTP, the hardware clock is automatically synchronized to the system clock every 11 minutes, and this command is useful only at boot time to get a reasonable initial system time. 
```

Since our testnodes are basically ephemeral now (reimaged before every job), my theory is that we'd want to make sure the hardware clock is set so that the machine gets a "reasonable initial system time."

Another goal is to make sure all our systems' hardware clocks are using UTC.  This isn't usually the case when they come from the manufacturer.

From the `hwclock` man page:

```
The  System  Time  is  the time that matters.  The Hardware Clock's basic purpose is to keep time when Linux is not running so that the System Clock can be initialized from it at boot.
```

PR #395, opened 2018-04-10
tools: Force time synchronization in FOG images

djgalloway:


djgalloway:
I may end up reverting this.  It's easiest to just merge it, run the Jenkins job, and revert if necessary (although I don't think I'll need to).

PR #400, opened 2018-05-16
Comment out vault password file path

djgalloway:
Having a blank/empty file is no longer supported in Ansible


vasukulkarni:
lets also please wait for 1 or 2 jobs to finish here then: http://pulpito.ceph.redhat.com/vasu-2018-05-16_15:01:20-smoke-luminous-distro-basic-bruuni/

djgalloway:
Just kidding.  Vasu's right.  This will break things in Octo.

PR #414, opened 2018-09-27
Support zypper for teuthology role

kshtsk:
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>

### DIFF
--- a/roles/paddles/tasks/main.yml
+++ b/roles/paddles/tasks/main.yml
@@ -25,6 +25,9 @@
   tags:
     - always
 
+- import_tasks: zypper_systems.yml
+  when: ansible_pkg_mgr == "zypper"
+
 - import_tasks: apt_systems.yml
   when: ansible_pkg_mgr == "apt"
 
@@ -46,3 +49,4 @@
 
 # Configure nginx as a reverse proxy
 - import_tasks: nginx.yml
+  when: not ansible_distribution is search("openSUSE")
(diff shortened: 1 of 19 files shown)


kshtsk:
@djgalloway since we paddles and pulpito are actually parts of teuthology, can we have these in one commit?

djgalloway:
teuthology is the testing tool (which can be run by itself).  paddles is just a database front end.  pulpito is just a Web UI front end for the database.

I know it's a pain but there's a ton of code here including new variables and no update to the docs.  Splitting into three commits based on the role would be the bare minimum I'd be comfortable reviewing.

kshtsk:
> teuthology is the testing tool (which can be run by itself). paddles is just a database front end. pulpito is just a Web UI front end for the database.
> 
> I know it's a pain but there's a ton of code here including new variables and no update to the docs. Splitting into three commits based on the role would be the bare minimum I'd be comfortable reviewing.

In ideal world probably they are separate tools, but reality is different. Try and schedule a suite with `teuthology-suite` command without paddles setup, and you'll see that workers will fail because they can't change the job status. Or try and run pulpito without paddles installed.
I understand your point, but I simply can't test the patch without having everything within different patches. It is possible but it is very painful.
If I rename the patch `Support zypper for teuthology, paddles, and pulpito role` will this work?

kshtsk:
@djgalloway #416 #417 #418

PR #418, opened 2018-10-03
Add zypper support for teuthology role

kshtsk:
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>

### DIFF
--- a/roles/teuthology/templates/teuthology.yaml
+++ b/roles/teuthology/templates/teuthology.yaml
@@ -4,7 +4,7 @@ results_server: {{ paddles_address }}
 results_ui_server: {{ pulpito_address }}
 results_email: {{ teuthology_results_email|default('null') }}
 results_sending_email: {{ teuthology_results_sending_email|default('null') }}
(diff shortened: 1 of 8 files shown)


roles/teuthology/templates/teuthology.yaml
```diff
@@ -4,7 +4,7 @@ results_server: {{ paddles_address }}
 results_ui_server: {{ pulpito_address }}
 results_email: {{ teuthology_results_email|default('null') }}
 results_sending_email: {{ teuthology_results_sending_email|default('null') }}
-lab_domain: {{ lab_domain }}
+lab_domain: {{ lab_domain|default('teuthology') }}
```
djgalloway: Could this go in https://github.com/ceph/ceph-cm-ansible/blob/master/roles/teuthology/defaults/main.yml instead please?

djgalloway: Actually, it's fine.

djgalloway:
Thanks for the PRs!

PR #422, opened 2018-11-16
Add RHEL8!

djgalloway:
### cobbler
- `libselinux-python, libsemanage-python, policycoreutils-python, ntp, python-jwt` are not available in RHEL8 so they're skipped in the `cephlab_packages_rhel` snippet.
- Replaced `wget` with `curl` in `snippets/cephlab_rc_local` since curl is universally available and installed by default.  At the time I first started working on these changes, wget wasn't packaged for RHEL8 yet.

### common
- `s/ansible_pkg_mgr/ansible_os_family` - `ansible_pkg_mgr` was skipping a lot of tasks because `ansible_pkg_mgr=dnf` for RHEL8.
- `s/yum:/package:` - `package` is a package-manager-agnostic module that will use dnf or yum.
- In `rhel-entitlements.yml`, the `Register with subscription-manager.` task would return 0 but there aren't any RHEL8 repos in RHSM yet.  So the play doesn't successfully identify the OS as Beta.  I added a few additional `failed_when` to ensure RHEL7 and RHEL8 Beta distros get identified so `beta_repos.yml` is called.
- There is no nagios or epel for RHEL8 yet so those tasks are skipped for now

### testnode
- `yum-complete-transaction`, `yum-utils` are not available in RHEL8.  Since we use FOG now, those tasks are irrelevant anyway because the yum cache is updated earlier in the play.
- CPAN installation fails on RHEL8

### DIFF
--- a/roles/cobbler/templates/snippets/cephlab_rc_local
+++ b/roles/cobbler/templates/snippets/cephlab_rc_local
@@ -3,7 +3,10 @@
 # Set proper location for firstboot ansible post-install trigger
 #set distro = $getVar('distro','').split("-")[0]
 #set distro_ver = $getVar('distro','').split("-")[1]
-#if $distro == 'Fedora' and int($distro_ver) >= 22
+#if $distro == 'RHEL'
+#set distro_ver = $distro_ver.split(".")[0]
+#end if
+#if ($distro == 'Fedora' and int($distro_ver) >= 22) or ($distro == 'RHEL' and int($distro_ver) >= 8)
 #set script = '/etc/rc.d/rc.local'
 #else if $distro == 'openSUSE'
 #set script = '/etc/init.d/boot.local'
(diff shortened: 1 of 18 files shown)


set_python_path.yml
```diff
@@ -0,0 +1,18 @@
+---
+# This will set ansible_python_interpreter to use python3
+# if the shell module fails (like it will on RHEL8 since
+# /usr/bin/python is no more).
```
zmc: Nice method of working around this!

roles/cobbler/templates/snippets/cephlab_rc_local
```diff
@@ -144,7 +147,7 @@ set +x
 echo -e "==================================\nInstructing Cobbler to run Ansible\n      Waiting for completion\n==================================" > /dev/console
 set -x
 # Run the post-install trigger a second time
-wget --timeout=1800 -t1 -O /dev/null "http://$http_server:$http_port/cblr/svc/op/trig/mode/post/system/$system_name" || true
+curl -s "http://$http_server:$http_port/cblr/svc/op/trig/mode/post/system/$system_name" -o /dev/null || true
```
zmc: Is the lack of timeout here going to be a potential issue? Would using `/usr/bin/timeout` be helpful?

PR #429, opened 2018-12-12
testnode: Remove DHCP config tasks

djgalloway:
Fixes: https://tracker.ceph.com/issues/37614


zmc:
It's safe to remove all of this just because bionic is having problems?

djgalloway:
I didn't see the point in adding additional unnecessary tasks to work around existing unnecessary tasks but we can close this in favor of https://github.com/ceph/ceph-cm-ansible/pull/431

PR #437, opened 2019-02-12
add GPG checking for RHEL 8

ktdreyer:
The RHEL 8 beta is GPG-signed, so we might as well verify signatures.


ktdreyer:
I got this wrong and it needs the beta key, not the GA key. https://github.com/ceph/ceph-cm-ansible/pull/457 is the follow-up.

PR #441, opened 2019-03-06
support opensuse package for nameserver role

kshtsk:


kshtsk:
@djgalloway could you please review?

PR #450, opened 2019-03-15
teuthology: make sure beanstalkd is running

kshtsk:
The beanstalkd service should be running before any
try to start workers up

Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@gmail.com>

### DIFF
--- a/roles/teuthology/tasks/main.yml
+++ b/roles/teuthology/tasks/main.yml
@@ -42,3 +42,9 @@
 - import_tasks: setup_log_access.yml
   tags:
     - logs
+
+- name: Enable and start beanstalkd
+  service:
+    name: beanstalkd
+    state: started
+    enabled: yes


kshtsk:
@djgalloway could you please review

roles/teuthology/tasks/main.yml
```diff
@@ -42,3 +42,11 @@
...
   tags:
     - logs
+
+- name: Enable and start beanstalkd
+  service:
+    name: beanstalkd
+    state: started
+    enabled: yes
+  tags:
+    - always
```
zmc: Why the `always` tag here?

kshtsk: I honestly don't know what for 'always' tag is used, it was just copy past from another service starting code in another role.
However I can say, that beanstalkd is a requirement for successful workers start or schedul request.

zmc: It causes the task to be run regardless of `--tags` or `--skip-tags`; that won't be useful here, I don't think

kshtsk: @zmc I've removed the tag.

PR #452, opened 2019-03-19
nsupdate_web: make sure nsupdate user group exists

kshtsk:
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@gmail.com>

### DIFF
--- a/roles/nsupdate_web/tasks/main.yml
+++ b/roles/nsupdate_web/tasks/main.yml
@@ -15,9 +15,16 @@
     state: latest
   with_items: "{{ packages }}"
 
+- name: Create nsupdate group
+  group:
+    name: "{{ nsupdate_web_user }}"
+    state: present
+    system: true
+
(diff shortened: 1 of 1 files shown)


kshtsk:
@djgalloway could you please review

djgalloway:
The ansible `user` module uses `useradd` to create users.  By default, `useradd` will create the group unless otherwise specified.  Is that not the case on OpenSUSE?

https://www.computerhope.com/unix/useradd.htm

kshtsk:
> The ansible `user` module uses `useradd` to create users. By default, `useradd` will create the group unless otherwise specified. Is that not the case on OpenSUSE?
> 
> https://www.computerhope.com/unix/useradd.htm

It appears that not, at least for Leap 42.3, however I can double check this for Leap 15.0 and Leap 15.1

djgalloway:
It can't hurt anything I guess.  Was just curious.

PR #464, opened 2019-08-14
Updating ISO Image for Ubuntu-16-04

akraitman:
The copy of the image wore too old to work with kickstart


djgalloway:
The copy of the kernel and initrd on cobbler were too old to work with the newer installation packages.

We were seeing "Loading libc6-udeb failed for unknown reasons. Aborting." during kickstart.

PR #467, opened 2019-08-20
Added rhev-1tb-ssd.yml to the tools dir

akraitman:
This playbook Is for creating /vm partition where the RHEV virtual machines will reside
And also for creating a 50G swap partition


djgalloway:
Can you allow the possibility for using an entire device as swap?  On the irvingi, /dev/sdb isn't used at all so I'd like to create a partition on them for swap.  If we created 50GB swapfiles on the root disk, they'd fill up quickly.

djgalloway:
BTW, looks great so far :+1:

djgalloway:
RHV is dead

PR #472, opened 2019-08-28
teuthology: Correct yaml syntax

djgalloway:


kshtsk:
For the record, the syntax was correct and is processed correctly on the systems with newer libcloud versions installed.

PR #479, opened 2019-10-25
roles/testnode/vars: install podman|docker newer distros

liewegas:
This is for ceph-daemon.

@rjfd  is this correct for opensuse?


rjfd:
@liewegas opensuse leap 42.1 is very old and does not have the `podman` package.

rjfd:
Maybe drop the `opensuse_leap_42.1.yml` file and just update the `opensuse_leap_15.0.yml`one

liewegas:
> Maybe drop the `opensuse_leap_42.1.yml` file and just update the `opensuse_leap_15.0.yml`one

updated!

PR #480, opened 2019-10-27
testnode: remove multipath

idryomov:
multipath can interfere with krbd tests, particularly with
krbd_udev_enumerate.yaml.  Much like udev, multipath opens block
devices behind the scenes, making the block device busy when it
shouldn't be.

None of the lab tests depend on multipath (it's not installed on
ubuntu), so just remove it rather than messing with /etc/multipath.conf
and restarting multipathd.  If the need for multipath comes up, rbd
devices would need to be blacklisted like so:

  blacklist {
          devnode "^rbd[0-9]+"
  }

Signed-off-by: Ilya Dryomov <idryomov@gmail.com>

### DIFF
--- a/roles/testnode/vars/apt_systems.yml
+++ b/roles/testnode/vars/apt_systems.yml
@@ -3,6 +3,10 @@ ntp_service_name: ntp
 ssh_service_name: ssh
 nfs_service: nfs-kernel-server
 
+packages_to_remove:
+  # multipath interferes with krbd tests
+  - multipath-tools
+
 ceph_packages_to_remove:
   - ceph
   - ceph-common
--- a/roles/testnode/vars/yum_systems.yml
+++ b/roles/testnode/vars/yum_systems.yml
@@ -2,6 +2,10 @@
 ntp_service_name: ntpd
 ssh_service_name: sshd
 
+packages_to_remove:
+  # multipath interferes with krbd tests
+  - device-mapper-multipath
+
 # ceph packages that we ensure do not exist
 ceph_packages_to_remove:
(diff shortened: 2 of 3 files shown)


djgalloway:
I've had trouble in the past with multipath devices on mira.  If a multipath disk is created, I am totally unable to install a new OS on that machine via Cobbler.  anaconda sees `/dev/mpatha` instead of `/dev/sda` and I've never been able to fix that.

This hasn't been much of an issue after moving to FOG so I've let it ride.

That's all sort of just context or FYI.

I think there are some ceph-disk tests in teuthology that create and/or test multipath devices.  I found https://tracker.ceph.com/issues/14460 and https://tracker.ceph.com/issues/11881 for example.

Also https://github.com/ceph/ceph/commit/0cf0e88e2922bfd283540d3c5c40a9ebc7613178

I want to be clear that I'm *all for* getting rid of multipath stuff as it has only caused me headaches but I fear this is something Ceph supports and needs to be tested and we'd be removing that ability here.

Can we at least get a test run using a few different suites?  I'm guessing ceph-disk specifically?

We can test this PR using
```
overrides:
  ansible.cephlab:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    branch: wip-suse
```

idryomov:
If someone still cares about ceph-disk suite, it's probably easy to fix, but it shouldn't block this PR.  Since it installs the multipath package by itself, it clearly doesn't expect it to come in the fog image or be installed by the testnode role, so we aren't removing the ability to test things that need to be tested.

PR #481, opened 2019-10-28
Adding rook-recovery role which can be used for deploying the rook je…

akraitman:
…nkins instance in the rook CI aws vpc, The deployied instance will have two ebs volumes were the jobs and the jobs configuration data is located

### DIFF
--- a/roles/rook-recovery/README.rst
+++ b/roles/rook-recovery/README.rst
@@ -1,0 +1,8 @@
+rook-recovery
+==========
+
+This role is used to deploy the rook jenkins instance in the rook CI aws vpc, The deployed instance will have two ebs volumes where the jobs and the jobs configuration data is located 
+
+Notes
++++++
+In order to make him available the to public network once the instance is deployed you will need to add him to the load balancing target group under the rook ci vpc
--- a/roles/rook-recovery/tasks/main.yml
+++ b/roles/rook-recovery/tasks/main.yml
@@ -1,0 +1,66 @@
+---
+  - name: Include secrets
+    include_vars: "{{ secrets_path | mandatory }}/aws.yaml"
+    no_log: true
+    tags:
+      - always
+
+  - name: Launch instance
+    ec2:
+      aws_access_key: "{{ aws_access_key }}"
+      aws_secret_key: "{{ aws_secret_key }}"
+      key_name: "{{ keypair }}"
+      group: "{{ security_group }}"
+      instance_type: "{{ instance_type }}"
+      image: "{{ image }}"
+      wait: true
+      region: "{{ region }}"
+      vpc_subnet_id: "{{ vpc_subnet_id }}" 
+      assign_public_ip: yes
+      instance_tags: "{{ aws_tags }}"
+      wait: yes
+    register: ec2_instances
+
+  - name: print ec2 facts
+    debug:
+      var: ec2_instances
+
+  - name: Set name tag for AWS instance
+    ec2_tag:
+      aws_access_key: "{{ aws_access_key }}"
+      aws_secret_key: "{{ aws_secret_key }}"
+      region: "{{ region }}"
+      resource: "{{ item.1.id }}"
+      tags:
+        Name: "{{ aws_tags.Name }}-{{ '%02d' | format(item.0 + 1) }}"
+    with_indexed_items: "{{ ec2_instances.instances }}"
+    loop_control:
+      label: "{{ item.1.id }} - {{ aws_tags.Name }}-{{ '%02d' | format(item.0 + 1) }}"
+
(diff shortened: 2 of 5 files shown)


roles/rook-recovery/README.rst
```diff
@@ -0,0 +1,8 @@
+rook-recovery
+==========
+
+This role is used to deploy the rook jenkins instance in the rook CI aws vpc, The deployied instance will have two ebs volumes were the jobs and the jobs configuration data is located
```
djgalloway: Typo

roles/rook-recovery/README.rst
```diff
@@ -0,0 +1,8 @@
+rook-recovery
+==========
+
+This role is used to deploy the rook jenkins instance in the rook CI aws vpc, The deployied instance will have two ebs volumes were the jobs and the jobs configuration data is located 
+
+Notes
++++++
+In order to make him avalible to public network once the instance is deployed you will need to add him to the load balancing target group under the rook ci vpc
```
djgalloway: Typo

roles/rook-recovery/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
+---
+  - name: Include secrets
+    include_vars: "{{ secrets_path | mandatory }}/aws.yaml"
+    no_log: true
+    tags:
+      - always
+
+  - name: Launch instance
+    ec2:
+       aws_access_key: "{{ aws_access_key }}"
```
djgalloway: Looks like there's one extra space under `ec2:` than there should be.

roles/rook-recovery/vars/main.yml
```diff
@@ -0,0 +1,13 @@
...
+instance_type: m4.large
+security_group: rook-jenkins-group
+image: ami-09c003703cafa9056
+region: us-east-1
+vpc_subnet_id: subnet-c72b609b
+instance_name: Recovery-Rook-Jenkins
+aws_tags:
+  Name: "{{ instance_name }}"
+  Application: "Jenkins"
+
```
djgalloway: Extra `\n`

djgalloway:
What is the source of the AMI?  Is it a generic Jenkins instance or is it a backup of the existing Jenkins instance?  Something else?

akraitman:
> What is the source of the AMI? Is it a generic Jenkins instance or is it a backup of the existing Jenkins instance? Something else?

It's stable backup of a Jenkins instance because we need the ebs volumes were the jobs and configuration data is located

djgalloway:
How feasible is it to have this role repopulate a plain/vanilla Jenkins instance using the existing Jenkinsfile from https://github.com/rook/rook/blob/master/Jenkinsfile ?

roles/rook-recovery/README.rst
akraitman: fixed

roles/rook-recovery/README.rst
akraitman: fixed

roles/rook-recovery/tasks/main.yml
akraitman: fixed

roles/rook-recovery/vars/main.yml
akraitman: fixed

akraitman:
> How feasible is it to have this role repopulate a plain/vanilla Jenkins instance using the existing Jenkinsfile from https://github.com/rook/rook/blob/master/Jenkinsfile ?

It's something i will need to investigate and asses because that would need more steps that would connect and mount the ebs volumes to that vanila instance before starting the jenkins container.
In any case the role would need to use the backup were the plugins and configuration data is located    so i am not sure if we will get a benefit in repopulating a plain/vanilla Jenkins instance over the way that the role is doing it now

roles/rook-recovery/README.rst
djgalloway: You need to `git add` all the files you fixed then `git commit --amend --no-edit` then `git push -f origin Rook-Recovery-Role` to push the fixes.

akraitman: I did it yesterday but i forgot doing git add, now it should be fine

roles/rook-recovery/README.rst
```diff
@@ -0,0 +1,8 @@
+rook-recovery
+==========
+
+This role is used to deploy the rook jenkins instance in the rook CI aws vpc, The deployed instance will have two ebs volumes were the jobs and the jobs configuration data is located
```
djgalloway: s/were/where

akraitman: Fixed, thank you

PR #482, opened 2019-11-05
users: Tasks to allow skipping of role if all users/keys are already up to date

djgalloway:
I'd like some input on this.

1) Is there a better way to do https://github.com/ceph/ceph-cm-ansible/compare/wip-skip-users?expand=1#diff-9ac36c3dbec0caf35aa4369ba60b54bdR20-R29

2) I couldn't find an ansible module to do https://github.com/ceph/ceph-cm-ansible/compare/wip-skip-users?expand=1#diff-9ac36c3dbec0caf35aa4369ba60b54bdR15 better than just `cat`ing the file.  Did I miss something?  `slurp:` didn't do what I wanted.

3) Am I missing any edge cases where this could go wrong?  I tried to get it so that if there are any problems with `git` locally or reading the sentinel file that the tasks would still run.

As long as we keep the FOG images up to date when we add new lab users, this would save a minute per testnode per job.

### DIFF
--- a/roles/users/README.rst
+++ b/roles/users/README.rst
@@ -88,6 +88,14 @@ A list of usernames whose access is to be revoked::
 
     revoked_users: []
 
+The users role writes a sentinel file, ``/keys-repo-sha1``, to indicate the sha1 of the keys repo when ceph-cm-ansible last ran.  If the sha1 in that file matches the current keys repo HEAD sha1, users tasks will be skipped unless you set ``force_users_update: True``::
+
+    force_users_update: False
+
+By default, the users and pubkeys should be updated.  A task in ``main.yml`` changes this to ``False`` if the machine's users and keys are already up to date (unless ``force_users_update: True``)::
+
+    perform_users_role: True
+
 Tags
 ++++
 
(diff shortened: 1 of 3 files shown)


djgalloway:
> I'd like some input on this.
> 
>     1. Is there a better way to do https://github.com/ceph/ceph-cm-ansible/compare/wip-skip-users?expand=1#diff-9ac36c3dbec0caf35aa4369ba60b54bdR20-R29
> 
>     2. I couldn't find an ansible module to do https://github.com/ceph/ceph-cm-ansible/compare/wip-skip-users?expand=1#diff-9ac36c3dbec0caf35aa4369ba60b54bdR15 better than just `cat`ing the file.  Did I miss something?  `slurp:` didn't do what I wanted.
> 
>     3. Am I missing any edge cases where this could go wrong?  I tried to get it so that if there are any problems with `git` locally or reading the sentinel file that the tasks would still run.
> 
> 
> As long as we keep the FOG images up to date when we add new lab users, this would save a minute per testnode per job.

I took care of all my concerns by defaulting to True and only setting  ``perform_users_role: False`` if the sentinel file and keys repo sha1 match OR ``force_users_update: True``

```
$ time ansible-playbook users.yml --limit smithi131* -e force_users_update=True
...
real  1m7.038s

$ time ansible-playbook users.yml --limit smithi131*
...
real  0m4.999s
```

PR #487, opened 2019-11-15
common: Update yum metadata from upstream epel and exclude mirrors

djgalloway:

### DIFF
--- a/roles/common/tasks/epel.yml
+++ b/roles/common/tasks/epel.yml
@@ -27,6 +27,21 @@
     - epel-mirrorlist
     - epel-testing-mirrorlist
 
-- name: Clean yum cache
-  shell: yum clean all
-  when: epel_repo is defined and epel_repo is changed
+# A few times a year, we'd see a bunch of 404 errors during yum transactions that involved epel mirrors.
+# It has to do with upstream EPEL being updated and mirrors lagging behind.
+- name: Temporarily comment mirrorlist so we can update yum cache
+  replace:
+    path: "/etc/yum.repos.d/{{ item.key }}.repo"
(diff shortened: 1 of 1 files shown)


roles/common/tasks/epel.yml
```diff
@@ -27,6 +27,21 @@
     - epel-mirrorlist
     - epel-testing-mirrorlist
 
-- name: Clean yum cache
-  shell: yum clean all
-  when: epel_repo is defined and epel_repo is changed
+# A few times a year, we'd see a bunch of 404 errors during yum transactions that involved epel mirrors.
+# It has to do with upstream EPEL being updated and mirrors lagging behind.
+- name: Temporarily comment mirrorlist so we can update yum cache
```
alfredodeza: you could use the `-C` flag to just use whatever is local in the machine and skip going to the network:

```
-C, --cacheonly
              Tells yum to run entirely from system cache; does not download or update metadata.  When this is used by a non-root user, yum will run entirely  from  user  cache  in  $TMPDIR.   This  option
              doesn't stop yum from updating user cache from system cache locally if the latter is newer (this is always done when running as a user).
```

djgalloway: Are you suggesting add `-C` to all yum transactions or just to the `yum makecache` below?

djgalloway:
I chatted with an EPEL maintainer and we can't figure out why this happens.

metadata A is on the system
metadata B is what's on the mirrors
metadata C is what's at the baseurl

`mirrorlist` always trumps `baseurl` but even if I comment `baseurl` and try to update cache, it still fails.

PR #516, opened 2019-12-23
testnode: el7: install podman-docker

liewegas:
Need this for el7 nodes too--anywhere we run cephadm.

### DIFF
--- a/roles/testnode/vars/centos_7.yml
+++ b/roles/testnode/vars/centos_7.yml
@@ -47,6 +47,8 @@ packages:
   - python-nose
   - mpich
   - podman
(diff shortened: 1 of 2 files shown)


djgalloway:
```
<sage> dgalloway: https://github.com/ceph/ceph-cm-ansible/pull/516
<dgalloway> sage, is that what it's called in el7?
<dgalloway> and did you check to make sure it's available in a standard RHEL repo?
<sage> well it's available on the centos 7.6 box i provisioned
<sage> dgalloway: not sure if rhel would be different 
<dgalloway> i'll check
```

Confirmed package is in rhel-7-server-extras-rpms

PR #537, opened 2020-02-27
Adding Fedora 31 support to the common role

akraitman:

### DIFF
--- a/roles/common/tasks/nagios.yml
+++ b/roles/common/tasks/nagios.yml
@@ -40,11 +40,11 @@
     - ansible_os_family == "RedHat"
     - have_epel.rc == 0
 
-- name: Install nrpe package and dependencies (Ubuntu)
+- name: Install nrpe package and dependencies (Ubuntu/Fedora)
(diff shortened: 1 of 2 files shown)


roles/common/defaults/main.yml
```diff
@@ -48,3 +48,7 @@ nagios_allowed_hosts: "127.0.0.1"
 nrpe_selinux_packages:
   - libsemanage-python
   - policycoreutils-python
+
+nrpe_selinux_fedora_packages:
```
djgalloway: These should get defined as `nrpe_selinux_packages` in `roles/common/vars/fedora_31.yml`.

djgalloway: These should get defined as `nrpe_selinux_packages` in `roles/common/vars/fedora_31.yml`.

djgalloway: These should get defined as `nrpe_selinux_packages` in `roles/common/vars/fedora_31.yml`.

PR #536, opened 2020-02-25
testnodes: add vars for opensuse_leap_15.2

smithfarm:
Fixes: https://tracker.ceph.com/issues/44292
Signed-off-by: Nathan Cutler <ncutler@suse.com>

PR #538, opened 2020-02-29
add opensuse_leap.15.2

liewegas:
Just a copy of 15.1!

Also updated cobber.yml...

Signed-off-by: Sage Weil <sage@redhat.com>

### DIFF
--- a/roles/testnode/vars/opensuse_leap_15.2.yml
+++ b/roles/testnode/vars/opensuse_leap_15.2.yml
@@ -1,0 +1,86 @@
+---
+# vars specific to OpenSuse Leap 15.2
+packages_to_remove:
+  - gettext-runtime-mini
+
+packages:
+  - lsb-release
+  - sysstat
+  - gdb
+  - make
+  - git
+  - python-configobj
+  # for running ceph
+  - libedit0
+#  - libboost_thread1_54_0
+  - libboost_thread1_66_0
+  - xfsprogs
+  - podman
+  - gptfdisk
+  - parted
+  - libgcrypt20
+  - fuse
+  - fuse-devel
+  - libfuse2
+  ###
+  # for ceph-deploy
+  - python-virtualenv
+  ###
+  - openssl
+  - libuuid1
+  - btrfsprogs
+  # used by workunits
+  - attr
+  - valgrind
+  - python-nose
+  - ant
+#  - iozone
+  ###
+  # used by the xfstests tasks
+  - libtool
+  - automake
+  - gettext-runtime
+  - libuuid-devel
+  - libacl-devel
+  - bc
+  - xfsdump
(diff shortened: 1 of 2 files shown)


liewegas:
@djgalloway is this right?

roles/testnode/vars/opensuse_leap_15.2.yml
```diff
@@ -0,0 +1,86 @@
...
+  # used by workunits
+  - attr
+  - valgrind
+  - python-nose
+  - ant
+#  - iozone
+  ###
+  # used by the xfstests tasks
+  - libtool
+  - automake
```
tchaikov: do we still need this?

roles/testnode/vars/opensuse_leap_15.2.yml
```diff
@@ -0,0 +1,86 @@
...
+  - gettext-runtime
```
tchaikov: we don't do i18n using gettext, right?

roles/testnode/vars/opensuse_leap_15.2.yml
```diff
@@ -0,0 +1,86 @@
...
+  - lsb-release
+  - sysstat
+  - gdb
+  - make
+  - git
+  - python-configobj
+  # for running ceph
+  - libedit0
+#  - libboost_thread1_54_0
+  - libboost_thread1_66_0
```
tchaikov: we don't use boost shipped by distro.

roles/testnode/vars/opensuse_leap_15.2.yml
liewegas: yes, for xfstests

liewegas:
this file is 100% identical to the 15.1 file

i'm not sure about the cobbler part?

djgalloway:
:disappointed: oops.. the vars file got deleted in https://github.com/ceph/ceph-cm-ansible/commit/659d0a5a17f0fe41b7178bd6b6bd29276fe517c7

@smithfarm had already taken care of this in https://github.com/ceph/ceph-cm-ansible/pull/536

There also needs to be an entry in https://github.com/ceph/ceph-cm-ansible/blob/master/roles/cobbler_profile/defaults/main.yml#L130

smithfarm:
> this file is 100% identical to the 15.1 file

The 15.1 file has Python 2 packages in it. I don't think we should be using those. 

Can't we just reinstate the one from #536 that got accidentally deleted?

djgalloway:
> Can't we just reinstate the one from #536 that got accidentally deleted?

+1 I'll take care of this

liewegas:
thanks!

djgalloway:
https://github.com/ceph/ceph-cm-ansible/pull/539

PR #540, opened 2020-03-03
Adding CentOS-8.1.1911-aarch64 to cobbler

akraitman:


akraitman:
> Also need an entry in https://github.com/ceph/ceph-cm-ansible/blob/master/cobbler.yml#L34. Maybe `centos8.1-aarch64` as the tag?

Right, now it should work

PR #541, opened 2020-03-03
testnode: use Python 3 in openSUSE Leap 15.2

smithfarm:
We should not be using Python 2 at all...

Signed-off-by: Nathan Cutler <ncutler@suse.com>

### DIFF
--- a/roles/testnode/vars/opensuse_leap_15.2.yml
+++ b/roles/testnode/vars/opensuse_leap_15.2.yml
@@ -2,18 +2,20 @@
 # vars specific to OpenSuse Leap 15.2
 packages_to_remove:
   - gettext-runtime-mini
+  - python
+  - python-base
 
 packages:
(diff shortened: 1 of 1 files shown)


smithfarm:
@djgalloway @liewegas

roles/testnode/vars/opensuse_leap_15.2.yml
```diff
@@ -2,18 +2,20 @@
...
   - gdb
   - make
+  - zypper
   - git
-#  - python-configobj
+  - python3-configobj
   # for running ceph
   - libedit0
-#  - libboost_thread1_54_0
-  - libboost_thread1_66_0
```
smithfarm: Note: `libboost_thread1_66_0` is a dependency of `zypper`, so it *will* be installed on the nodes, but dropping it here since it's not needed for ceph.

PR #542, opened 2020-03-10
testnode: Use mirrorlists for CentOS 8 repos

djgalloway:
Fixes ``Failed to download metadata for repo 'extras'`` errors.

Signed-off-by: David Galloway <dgallowa@redhat.com>

PR #544, opened 2020-03-13
common: Don't use baseurl for epel at all

djgalloway:
Despite having a mirrorlist present in the repo file, if contacting the baseurl fails, the yum transaction fails which causes our jobs to fail.  Let's try *just* using a mirrorlist.

See https://github.com/ceph/ceph-cm-ansible/pull/542#issuecomment-598924481


liewegas:
still happening... :/  /a/sage-2020-03-14_14:05:42-rados:cephadm-wip-sage3-testing-2020-03-14-0747-distro-basic-smithi/4855720

djgalloway:
Attempting another fix now.

PR #555, opened 2020-03-23
Fixing the nameserver role to support rhel6

akraitman:

### DIFF
--- a/roles/common/tasks/nagios.yml
+++ b/roles/common/tasks/nagios.yml
@@ -82,7 +82,7 @@
   register: firewalld
   ignore_errors: true
   no_log: true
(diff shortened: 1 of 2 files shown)


roles/common/tasks/nagios.yml
```diff
@@ -82,15 +82,18 @@
   register: firewalld
   ignore_errors: true
   no_log: true
-  when: ansible_os_family == "RedHat"
+  when: ansible_os_family == "RedHat" and ansible_distribution_major_version|int >= "7"
```
djgalloway: When you do `and ansible_distribution_major_version|int`, the comparison (`7` in this case) shouldn't be double-quoted because it's comparing a number with a string.

PR #566, opened 2020-04-28
common: Define release with subscription-manager

djgalloway:
See comment


ktdreyer:
FYI I think https://bugzilla.redhat.com/show_bug.cgi?id=1286621 means that this is only supported in conjunction with Satellite.

djgalloway:
> FYI I think https://bugzilla.redhat.com/show_bug.cgi?id=1286621 means that this is only supported in conjunction with Satellite.

Yeah, I reverted it anyway because it broke things horribly: https://github.com/ceph/ceph-sepia-secrets/pull/503

PR #589, opened 2020-11-19
common: Support for container mirror CA certs

djgalloway:


dmick (approved):
Looks right.  Of course consumers still have to be convinced to use the mirror when pulling.

PR #612, opened 2021-03-03
testnode: libgcrypt needs to be updated to match QEMU build

dillaman:
See https://bugzilla.redhat.com/show_bug.cgi?id=1828681

Signed-off-by: Jason Dillaman <dillaman@redhat.com>

### DIFF
--- a/roles/testnode/vars/centos_8.yml
+++ b/roles/testnode/vars/centos_8.yml
@@ -35,6 +35,9 @@ yum_mirrorlists:
   - CentOS-Base-mirrorlist
   - CentOS-Extras-mirrorlist
 
+packages_to_upgrade:
+  - libgcrypt # explicitly tied to qemu build
+
 packages:
   # for package-cleanup
   - dnf-utils


dillaman:
See [1] for an example of what just started appearing on CentOS 8.2 builds. Due to this BZ [2], when the latest available QEMU is installed on a test node, it fails with a "Unable to initialize gcrypt" error because of an invalid check that compares the build-time version of libgcrypt with the runtime version. Since libgcrypt wasn't getting upgraded when a newer QEMU is installed, it's causing all QEMU tests to break.

[1] http://qa-proxy.ceph.com/teuthology/jdillaman-2021-03-02_19:09:33-rbd-wip-jd-testing-distro-basic-smithi/5928437/teuthology.log
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1828681

djgalloway (approved):

dillaman:
@djgalloway Thx! Looks like QEMU is happy again under CentOS 8.2 [1]

[1] http://qa-proxy.ceph.com/teuthology/jdillaman-2021-03-03_14:49:58-rbd-wip-jd-testing-testing-basic-smithi/5930859/teuthology.log

PR #613, opened 2021-03-03
testnode: Install collectl without recommended packages

djgalloway:
collectl has a bunch of Gnome Desktop Environment packages that are recommended.  So on Ubuntu 20.04 at least, a desktop environment was getting installed and would eventually go to sleep.


dmick (approved):
Looks right to me.

PR #617, opened 2021-04-01
Adding the releasever variable to rhel7.9 testnodes

akraitman:
Fixes: https://tracker.ceph.com/issues/49771

### DIFF
--- a/roles/testnode/tasks/yum/repos.yml
+++ b/roles/testnode/tasks/yum/repos.yml
@@ -48,3 +48,10 @@
   when: (repo_file is defined and repo_file is changed) or
         (gpg_keys is defined and gpg_keys is changed) or
         (version_repo_file is defined and version_repo_file is changed)
+
+# set the releasever cause without it rhel-7-server-rpms repo fails on rhel7.9 machines https://tracker.ceph.com/issues/49771
+- name: set the releasevar on rhel
+  copy:
+    content: "{{ ansible_lsb.release }}"
+    dest: /etc/yum/vars/releasever
+  when: ansible_distribution == "RedHat"


djgalloway (changes requested):
A few things:

- These tasks only get pulled on on RHEL so the `ansible_distribution == "RedHat"` is redundant.
- Use the `ansible_lsb.release` var and make it version-independent
- Use a module.  Maybe `copy`

```
copy:
  content: "{{ ansible_lsb.release }}"
  dest: /etc/yum/vars/releasever
```

Make sure you test this by running the entire `cephlab.yml` playbook on a RHEL7.9 and RHEL8.3 testnode.

akraitman:
> A few things:
> 
> * These tasks only get pulled on on RHEL so the `ansible_distribution == "RedHat"` is redundant.
> * Use the `ansible_lsb.release` var and make it version-independent
> * Use a module.  Maybe `copy`
> 
> ```
> copy:
>   content: "{{ ansible_lsb.release }}"
>   dest: /etc/yum/vars/releasever
> ```
> 
> Make sure you test this by running the entire `cephlab.yml` playbook on a RHEL7.9 and RHEL8.3 testnode.

I placed it in testnode/tasks/yum/repos.yml instead since i saw it has to run before the "Install packages" task , tested on rhel7.9 and on rhel8.3

djgalloway:
> I placed it in testnode/tasks/yum/repos.yml instead since i saw it has to run before the "Install packages" task , tested on rhel7.9 and on rhel8.3

Okay, well in that case, you will want `when: ansible_distribution == "RedHat"`.  Please also add a comment above the task explaining what it's for.

Let's try removing the `ansible_distribution_version == "7.9"` and see what happens on RHEL7.8 and RHEL8.x.

akraitman:
> > I placed it in testnode/tasks/yum/repos.yml instead since i saw it has to run before the "Install packages" task
> 
> Okay, well in that case, you will want `when: ansible_distribution == "RedHat"`. Please also add a comment above the task explaining what it's for.
> 
> Let's try removing the `ansible_distribution_version == "7.9"` and see what happens on RHEL7.8 and RHEL8.x.

Okay, did this change and tested it on rhel7.9 and rhel8.3

djgalloway:
Edit the comment to include the link to the tracker then we're good.

djgalloway (approved):

PR #619, opened 2021-04-15
teuthology.yaml: add vip configuration

liewegas:
Signed-off-by: Sage Weil <sage@newdream.net>

### DIFF
--- a/roles/teuthology/templates/teuthology.yaml
+++ b/roles/teuthology/templates/teuthology.yaml
@@ -14,3 +14,7 @@ archive_base: {{ archive_base }}
 ceph_git_base_url: {{ teuthology_ceph_git_base_url }}
 queue_host: localhost
 queue_port: 11300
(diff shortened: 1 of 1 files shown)


liewegas:
@djgalloway ping?  i've manually edited the teuthworker .teuthology.yaml, but presumably this should also be updated?

liewegas:
@jdurgin

djgalloway:
It should but we shouldn't hardcode it in the ceph-cm-ansible.git version.  It should be parameterized.

So blank definitions here: https://github.com/ceph/ceph-cm-ansible/blob/master/roles/teuthology/defaults/main.yml

And our (Sepia's) subnets should be defined here: https://github.com/ceph/ceph-sepia-secrets/blob/master/ansible/inventory/group_vars/teuthology.yml

djgalloway:
Closing as stale

PR #621, opened 2021-05-13
cobbler: Support CentOS Stream

djgalloway:
We have to do it this way for a number of reasons

1. The vmlinuz and initrd in the CentOS Stream ISO aren't compatible. https://bugs.centos.org/view.php?id=18188
2. The version of Cobbler we're running does not recognize CentOS Stream ISO without hacking the distro_signatures file
3. Probably most importantly, when using the CentOS Stream ISO, after the installation is finished, we bizarrely end up with CentOS 8.3...  Not Stream.

### DIFF
--- a/roles/cobbler_profile/tasks/import_stream_profile.yml
+++ b/roles/cobbler_profile/tasks/import_stream_profile.yml
@@ -1,0 +1,26 @@
+---
+- name: "Extract distro name and major version from {{ distro_name }}"
+  set_fact:
+    distro_and_version: "{{ distro_name.split('.')[0] }}"
+
+- name: "Extract distro name from {{ distro_name }}"
+  set_fact:
+    stream_distro_name: "{{ distro_name.split('-')[0] }}"
+
+- name: "Extract the major version number from {{ distro_and_version }}"
+  set_fact:
+    stream_distro_version: "{{ distro_and_version.split('-')[1] }}"
+
+- name: "Get the latest non-Stream profile that matches this {{ stream_distro_name }} Stream distro version (e.g., CentOS-8.3-x86_64)"
+  shell: "cobbler profile list | grep {{ distro_and_version }} | grep -v 'stream\\|arm\\|aarch\\|stock' | sort -V | tail -n 1 | xargs"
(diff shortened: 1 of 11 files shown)


roles/cobbler_profile/tasks/import_stream_profile.yml
```diff
@@ -0,0 +1,26 @@
...
+
+- name: "Extract the major version number from {{ distro_and_version }}"
+  set_fact:
+    stream_distro_version: "{{ distro_and_version.split('-')[1] }}"
+
+- name: "Get the latest non-Stream profile that matches this {{ stream_distro_name }} Stream distro version (e.g., CentOS-8.3-x86_64)"
+  shell: "cobbler profile list | grep {{ distro_and_version }} | grep -v 'stream\\|arm\\|aarch\\|stock' | sort -V | tail -n 1 | xargs"
+  register: latest_non_stream_profile
+
+# See commit message for why we do it this way
```
zmc, line 18: Almost got me. There's no commit message with an explanation.

zmc (changes requested):
One comment that should be addressed. Also, a non-blocking suggestion:

I see you use `{% if 'Stream' in ansible_lsb.description %}-stream{% endif %}` in a _lot_ of places; would it make sense to use `set_fact` to set a variable to that value e.g. `stream_prefix` then simply use `{stream_prefix}` in the templates? Its value would either be `'-stream'` or `''`

roles/cobbler_profile/tasks/import_stream_profile.yml
djgalloway: Oh sonofa

djgalloway:
> One comment that should be addressed. Also, a non-blocking suggestion:
> 
> I see you use `{% if 'Stream' in ansible_lsb.description %}-stream{% endif %}` in a _lot_ of places; would it make sense to use `set_fact` to set a variable to that value e.g. `stream_prefix` then simply use `{stream_prefix}` in the templates? Its value would either be `'-stream'` or `''`

Yep.  Good suggestion.  I will implement that!

akraitman (approved):

PR #633, opened 2021-06-16
container-host: podman-docker and docker (from EPEL) conflict

djgalloway:
Which totally makes sense. :eyeroll:


dmick (commented):
Kind of.  podman-docker is "don't use docker, use podman instead" and actually installs a docker redirection shell script.  Anyway, LGTM if we're saying "we don't use podman on c7", which I can easily imagine.

dmick (approved):

PR #644, opened 2021-08-17
Adding jenkins ocp4 jobs restore playbook

akraitman:

### DIFF
--- a/tools/downstream-jenkins-sync-jobs.yml
+++ b/tools/downstream-jenkins-sync-jobs.yml
@@ -1,0 +1,51 @@
+---
+# This playbook is used to sync the jenkins jobs from one ocp pod folder to another pod folder
+# Usage:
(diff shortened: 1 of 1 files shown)


tools/downstream-jenkins-sync-jobs.yml
```diff
@@ -0,0 +1,53 @@
...
+    register: oc_whoami
+
+  - name: Fail if oc user is not connected
+    fail:
+      msg: "Please login to the ocp cluster by running oc login"
+    when: oc_whoami.rc != 0
+
+  - name: Create local directory
+    file:
+      path: "{{ local_tmp_folder }}"
```
djgalloway, line 41: Maybe just have the playbook create this directory instead of having to manually create and then pass it as a variable?

`mktemp -d`

akraitman: Changed it so that now it creates a tmp dir without passing a variable

djgalloway: :+1: good work!

djgalloway (approved):

tools/downstream-jenkins-sync-jobs.yml
akraitman: Thanks man!

PR #646, opened 2021-08-19
roles/testnode/templates/ssh: disable GSSAPIAuthentication for RHEL8 and Centos8

melissa-kun-li:
Disable GSSAPIAuthentication on the RHEL8 and Centos8 testnodes for this PR: https://github.com/ceph/ceph/pull/42051

Fixes: https://tracker.ceph.com/issues/44676
Signed-off-by: Melissa Li <li.melissa.kun@gmail.com>

### DIFF
--- a/roles/testnode/templates/ssh/sshd_config_centos_8
+++ b/roles/testnode/templates/ssh/sshd_config_centos_8
@@ -18,8 +18,8 @@ PasswordAuthentication yes
 ChallengeResponseAuthentication no
 
 # GSSAPI options
-GSSAPIAuthentication yes
-GSSAPICleanupCredentials yes
+GSSAPIAuthentication no
+#GSSAPICleanupCredentials yes
 
 UsePAM yes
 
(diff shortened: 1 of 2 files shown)


djgalloway:
Aren't customers/users going to run into issues if they have GSSAPI enabled?  Or is it just the unit test that is failing?

tchaikov:
https://github.com/ceph/ceph/pull/42051/files#r693283223

melissa-kun-li:
> Aren't customers/users going to run into issues if they have GSSAPI enabled? Or is it just the unit test that is failing?

https://github.com/ceph/ceph/pull/42051#issuecomment-903036575 Hmm, right it's probably not the best to disable it here. Instead of disabling it here, I'm disabling the GSSAPI auth within the code in the asyncssh.connect call (only affects cephadm which doesn't need GSSAPI), so users shouldn't be affected if they have `GSSAPIAuthentication yes` in their ssh config. I'll close this

PR #648, opened 2021-09-23
DNM: paddles: Enable containerized deployment

zmc:
This is not entirely finished


zmc:
@kshtsk You may want to take a look at this

roles/paddles/defaults/main.yml
```diff
@@ -1,9 +1,17 @@
 ---
 paddles_user: paddles
+paddles_db_user: paddles
+paddles_port: 8080
+paddles_statsd_host: ""
+paddles_statsd_prefix: ""
+paddles_sentry_dsn: ""
+
+paddles_containerized: false
+paddles_container_image: quay.io/ceph-infra/paddles:latest
```
kshtsk, line 10: I guess this image is supposed to be the latest and greatest some how tested before. If so, do we have somewhere a note how can I test paddles image from my local source code before submitting as a PR?

zmc: Great question. I will write something up to help with that this week.

The latest tag is an alias to master, which is a common convention. If you look [here](https://quay.io/repository/ceph-infra/paddles?tab=tags), you can see some other tags. Each branch of this repo will have containers built automatically. For someone w/o push access to this repo, they'd have to build their own image - and it's possibly to do exactly what I did on quay.io for a forked repo under a personal account for free as well.

roles/paddles/tasks/main.yml
```diff
@@ -40,13 +46,23 @@
...
 # Configure the system to run paddles as a daemon
 - import_tasks: setup_service.yml
+  when: not paddles_containerized
 
 # Configure nginx as a reverse proxy
 - import_tasks: nginx.yml
-  when: not ansible_distribution is search("openSUSE")
+  when:
+    - not ansible_distribution is search("openSUSE")
+    - not paddles_containerized
```
djgalloway, line 68: The reverse proxy is no longer needed if it's containerized?  How's that work?

roles/users/tasks/main.yml
```diff
@@ -11,12 +11,16 @@
   delay: 10
   # perform_users_role is True by default so no need to fail the play if there's an error.
   ignore_errors: true
+  tags:
+    - pubkeys
```
djgalloway, line 15: Did you mean to commit the changes in this file?

roles/paddles/tasks/main.yml
zmc: I'm glad you caught this; in theory we could ditch nginx in favor just letting docker handle the load balancing, but _I forgot to push a commit that removes that line_ because I decided against doing that for the time being

roles/users/tasks/main.yml
zmc: I did yeah; I was having problems getting `--skip-tags user,pubkeys` to work properly without this

djgalloway: Weird.  What was the full command you were running?

zmc: Something like `ansible-playbook -v -i ~/dev/ceph-sepia-secrets/ansible/inventory/sepia -l 'paddles.front.*' ./paddles.yml --skip-tags user,pubkeys`

djgalloway:
Oh, wait, what about docs?

PR #651, opened 2021-10-05
common: Increase NRPE command timeout

djgalloway:


djgalloway:
We no longer use Nagios

PR #665, opened 2022-01-19
nameserver: Check for and print IP collisions

djgalloway:


zmc (approved):
I think this looks good. Does a human have to happen to notice this output in the logs though?

djgalloway:
> I think this looks good. Does a human have to happen to notice this output in the logs though?

Yes but it's the last thing that gets printed.

PR #666, opened 2022-02-02
Fix non-stream to stream conversion

djgalloway:
The previous method no longer works

### DIFF
--- a/roles/cobbler_profile/tasks/import_distro.yml
+++ b/roles/cobbler_profile/tasks/import_distro.yml
@@ -8,11 +8,8 @@
   debug: var=distro_name
 
 - name: Load extra_distros from secrets
-  # The jinja2 templating API allows you to update one dict with another,
-  # but it does so in-place without returning any dict. The first few
-  # blobs do the updating; the last actually returns the result.
   set_fact:
-    distros: "{% for distro in distros.keys() + extra_distros.keys() %}{{ distros[distro].update(extra_distros[distro]|default({})) }}{% endfor %}{{ distros }}"
+    distros: "{{ distros|combine(extra_distros, recursive=True) }}"
 
 - name: Find distro settings
   set_fact:
--- a/tools/convert-to-centos-stream.yml
+++ b/tools/convert-to-centos-stream.yml
@@ -8,6 +8,39 @@
   gather_facts: true
   tasks:
 
+  - name: List repo files
+    find:
+      paths: /etc/yum.repos.d/
+      file_type: file
+      patterns: 'CentOS-Linux-*.repo'
+    register: pre_stream_repo_files
+    when: ansible_distribution == 'CentOS'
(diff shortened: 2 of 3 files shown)


tools/convert-to-centos-stream.yml
```diff
@@ -8,6 +8,39 @@
...
+      file_type: file
+      patterns: 'CentOS-Linux-*.repo'
+    register: pre_stream_repo_files
+    when: ansible_distribution == 'CentOS'
+
+  - name: Remove all mirrorlists
+    replace:
+      path: "{{ item.path }}"
+      regexp: '^mirrorlist=.*'
+#      replace: 'release=$releasever-stream&'
```
zmc, line 23: Mistake?

djgalloway: Shoot.  Yeah.

roles/cobbler_profile/tasks/import_distro.yml
```diff
@@ -8,11 +8,8 @@
   debug: var=distro_name
 
 - name: Load extra_distros from secrets
-  # The jinja2 templating API allows you to update one dict with another,
-  # but it does so in-place without returning any dict. The first few
-  # blobs do the updating; the last actually returns the result.
   set_fact:
-    distros: "{% for distro in distros.keys() + extra_distros.keys() %}{{ distros[distro].update(extra_distros[distro]|default({})) }}{% endfor %}{{ distros }}"
+    distros: "{{ distros|combine(extra_distros, recursive=True) }}"
```
zmc, line 12: this is so much nicer

tools/convert-to-centos-stream.yml
```diff
@@ -8,6 +8,39 @@
...
+      regexp: '^\s*#*\s*(baseurl=.*)'
+      replace: '\1'
+    with_items: "{{ pre_stream_repo_files.files }}"
+    when: ansible_distribution == 'CentOS'
+
+  - name: Point baseurls to archive server
+    replace:
+      path: "{{ item.path }}"
+      regexp: 'mirror.centos.org/\$contentdir/\$releasever'
+      replace: 'vault.centos.org/8.5.2111'
```
zmc, line 40: This feels like something that could go stale without us really noticing; should it maybe be abstracted to a variable?

zmc (approved):
Looks good! One concern that isn't a blocker

tools/convert-to-centos-stream.yml
djgalloway: Eh.  My thinking was

1. vault.centos.org is supposed to be the vault.  Like it doesn't change.
2. There will never be a newer version of CentOS 8.  8.5.2111 is the last one.

zmc: Oh, of course! 🤦

PR #667, opened 2022-02-03
Revert "testnode: Add Apache Arrow repo"

cbodley:
This reverts commit 3f960fe1b6957bdfb551bd4e53ad275d0d446855.

the new plan is to submodule the things we needed from this repo, so it's no longer necessary. and since centos 8 eol, it looks like apache may have removed the `/centos/8/` repos so this is now causing teuthology failures


neha-ojha (approved):

cbodley:
cc @galsalomon66

PR #670, opened 2022-02-10
testnode: Alternate method of detecting Stream (for Openstack instances)

djgalloway:
Apparently the CentOS Stream Openstack images work despite the PXE images not working.

These changes are needed in order to run the testnode role against CentOS 8 Stream instances in Openstack.


kshtsk:
maybe we can add lvm2 to packages for centos 8.stream as separate commit to this PR?

djgalloway:
> maybe we can add lvm2 to packages for centos 8.stream as separate commit to this PR?

Done

kshtsk:
nice, I would approve this if I had this right for the repo :-D

PR #671, opened 2022-03-09
testnode: Filter ansible_mounts with regex

zmc:
This fixes an issue Seagate was hitting

### DIFF
--- a/roles/testnode/tasks/zap_disks.yml
+++ b/roles/testnode/tasks/zap_disks.yml
@@ -32,15 +32,7 @@
   with_items: "{{ ansible_mounts }}"
   when:
     - item.mount != '/' and
-      item.mount != '/var' and
-      item.mount != '/tmp' and
-      item.mount != '/root' and
-      item.mount != '/home' and
-      '"/boot" not in item.mount' and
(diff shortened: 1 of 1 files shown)


djgalloway:
Oooo I like this.  Can we combine all those options into one jinja2 regex?

roles/testnode/tasks/zap_disks.yml
```diff
@@ -32,7 +32,7 @@
   with_items: "{{ ansible_mounts }}"
   when:
     - item.mount != '/' and
-      item.mount != '/var' and
+      not item.mount.startswith('/var') and
```
djgalloway, line 35: Would `startswith('/var|/tmp|/root|/etc')` work?

zmc: I don't see why not, should I make that change?

djgalloway: Yeah!

zmc: Looks like I could replace all but the first condition (`!= /`) with a regex. Sound good?

djgalloway (approved):

PR #672, opened 2022-03-09
Moving toward supporting containerized testnodes

zmc:


roles/testnode/tasks/apt/packages.yml
```diff
@@ -12,7 +12,9 @@
     name: "{{ packages_to_upgrade|list }}"
     state: latest
     force: yes
-  when: packages_to_upgrade|length > 0
+  when:
+  - packages_to_upgrade|length > 0
+  - not containerized_node
```
djgalloway, line 17: Why can't we upgrade packages in containers?  What was the problem?

roles/testnode/tasks/apt_systems.yml
```diff
@@ -38,18 +38,24 @@
...
+    - not containerized_node
 
 - name: Enable xattr for this boot.
   command:
     mount -o remount,user_xattr /
   when: add_user_xattr is defined and
         add_user_xattr is changed
 
-- name: Ensure fuse group exists.
+- name: Ensure fuse, kvm and disk groups exist.
```
djgalloway, line 51: This tested on baremetal?

roles/testnode/tasks/ntp.yml
```diff
@@ -3,7 +3,7 @@
   yum:
     name: ntp
     state: present
-  when: ansible_pkg_mgr  == "yum"
+  when: ansible_pkg_mgr  == "yum" or ansible_pkg_mgr == "dnf"
```
djgalloway, line 6: Just change to `ansible_os_family == "RedHat"`

roles/testnode/tasks/apt_systems.yml
zmc: No but I can do that

roles/testnode/tasks/apt/packages.yml
zmc: Ah, this can be left out. The "problem" is that for the workflow I'm testing, it brings a <2min ansible run up to ~9min. But I think I can also just skip the packages tag.

roles/testnode/tasks/apt_systems.yml
zmc: Tested here https://pulpito.ceph.com/zack-2022-03-11_01:40:50-orch:cephadm:smoke-master-distro-default-smithi/

roles/testnode/tasks/ntp.yml
zmc: I ended up dropping the commit

djgalloway (approved):

PR #673, opened 2022-03-31
Move secondary NIC tasks to common role

djgalloway:
I want to use them for the new LRC nodes


zmc (approved):
Love the fine-grained commits; they make reviews like this trivial

PR #675, opened 2022-04-05
testnode: Fix CentOS 8 Stream packages list

djgalloway:
I guess we were relying on a bug for the past few months. What would happen is centos_8.yml would get pulled in and the `packages` list got populated. Then later when we detect this is CentOS Stream, centos_8_stream.yml would get pulled in which overrode the original `packages` list.

### DIFF
--- a/roles/testnode/vars/centos_8_stream.yml
+++ b/roles/testnode/vars/centos_8_stream.yml
@@ -1,8 +1,66 @@
 ---
 # vars specific to centos stream version 8.x
 
+common_yum_repos:
+  lab-extras:
+    name: "lab-extras"
+    baseurl: "http://{{ mirror_host }}/lab-extras/8/"
+    enabled: 1
+    gpgcheck: 0
+
 packages_to_upgrade:
   - systemd
+  - libgcrypt
 
 packages:
+  - redhat-lsb-core
+  # for package-cleanup
+  - dnf-utils
+  - sysstat
+  - libedit
+  - boost-thread
+  - xfsprogs
+  - gdisk
+  - parted
+  - libgcrypt
+  - fuse-libs
+  - openssl
+  - libuuid
+  - podman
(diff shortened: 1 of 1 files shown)


roles/testnode/vars/centos_8_stream.yml
```diff
@@ -1,8 +1,66 @@
 ---
 # vars specific to centos stream version 8.x
 
+common_yum_repos:
+  lab-extras:
+    name: "lab-extras"
+    baseurl: "http://{{ mirror_host }}/lab-extras/8/"
+    enabled: 1
+    gpgcheck: 0
```
idryomov, line 9: It looks like you brought over `common_yum_repos` stanza from `centos_8.yml` but not `nfs_service` or `ntp_service_name`?  If we are still relying on both `centos_8.yml` and `centos_8_stream.yml` being applied for Stream, why bringing over `common_yum_repos` (which is exactly the same) was neccesary?

Perhaps we should get rid of this fragility altogether and have `centos_8.yml` apply strictly to non-Stream and `centos_8_stream.yml` apply strictly to Stream?

djgalloway: > It looks like you brought over `common_yum_repos` stanza from `centos_8.yml` but not `nfs_service` or `ntp_service_name`? If we are still relying on both `centos_8.yml` and `centos_8_stream.yml` being applied for Stream, why bringing over `common_yum_repos` (which is exactly the same) was neccesary?

Good point.

> Perhaps we should get rid of this fragility altogether and have `centos_8.yml` apply strictly to non-Stream and `centos_8_stream.yml` apply strictly to Stream?

Because centos_8.yml also applies to CentOS 8 Stream.  Only `packages: []` and `packages_to_upgrade: []` need to be overridden.

djgalloway: https://github.com/ceph/ceph-cm-ansible/pull/676

PR #683, opened 2022-06-16
testnode: Add missing parenthesis to conditional

zmc:


zmc:
fun fact: `ansible-lint` didn't catch this!

PR #686, opened 2022-06-28
tools: Add a couple workstation scripts

djgalloway:


djgalloway:
Self-merging as these are just some scripts for workstation use.

PR #712, opened 2022-12-25
Using the cpanm ansible module

akraitman:
Using the cpanm ansible module instead of using the "command" module since we are seeing "Proceed nonetheless? [no] no', 'Aborted." during host re-imaging which can be solved using this module that does a clean installation in this case
http://pulpito.front.sepia.ceph.com/yuvalif-2022-12-20_21:29:08-rgw-wip-yuval-fix-58167-distro-default-smithi/7122795/

### DIFF
--- a/roles/testnode/tasks/cpan.yml
+++ b/roles/testnode/tasks/cpan.yml
@@ -46,6 +46,7 @@
   changed_when: false
 
 - name: "Install Amazon::S3."
-  command: "cpan  Amazon::S3"
+  cpanm:
+    name: "Amazon::S3"
   when: cpan_check is defined and
         cpan_check.rc != 0


idryomov (commented):
This seems straightforward enough to me but it would be nice to explain why the direct `cpan` invocation stopped working or at least provide a link to the tracker ticket.

dmick:
What @idryomov said.  Both the commit message and the PR should have a motivation, even if it's just a link to a tracker (although a tracker issue isn't *required* IMO)

akraitman:
> This seems straightforward enough to me but it would be nice to explain why the direct `cpan` invocation stopped working or at least provide a link to the tracker ticket.

http://pulpito.front.sepia.ceph.com/yuvalif-2022-12-20_21:29:08-rgw-wip-yuval-fix-58167-distro-default-smithi/7122795/

dmick:
@akraitman: please explain here what happened to break the cpan invocation and why adding cpanm fixes the problem.

PR #714, opened 2023-01-06
auth-openvpn: port to python3 for new gw server, futures

dmick:
I chose to keep everything in binary throughout.  One could also choose string, but this seemed safer for international characters.

Signed-off-by: Dan Mick <dmick@redhat.com>

### DIFF
--- a/roles/gateway/templates/auth-openvpn
+++ b/roles/gateway/templates/auth-openvpn
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 import hashlib
 import logging
@@ -15,13 +15,13 @@ def authenticate():
     time.sleep(1)
 
     path = sys.argv[1]
-    with file(path, 'rb') as f:
+    with open(path, 'rb') as f:
         user = f.readline(8192)
-        assert user.endswith('\n')
+        assert user.endswith(b'\n')
         user = user[:-1]
         assert user
         secret = f.readline(8192)
-        assert secret.endswith('\n')
+        assert secret.endswith(b'\n')
         secret = secret[:-1]
         assert secret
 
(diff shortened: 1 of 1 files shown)


roles/gateway/templates/auth-openvpn
```diff
@@ -36,16 +36,16 @@ def authenticate():
...
+        with open('/etc/openvpn/cephco-sepia.data/users', 'rb') as f:
             for line in f:
-                assert line.endswith('\n')
+                assert line.endswith(b'\n')
                 line = line[:-1]
-                if line.startswith("#") or len(line) == 0:
+                if line.startswith(b'#') or len(line) == 0:
                     continue
-                (username, salt, correct) = line.split(' ', 2)
+                (username, salt, correct) = line.split(b' ', 2)
```
zmc, line 48: github won't let me comment on line 54 because it's not near enough to a line that was modified in this PR, but `salt` will need to be `bytes` to prevent `inner.update(salt)` from throwing a `TypeError`

roles/gateway/templates/auth-openvpn
```diff
@@ -36,16 +36,16 @@ def authenticate():
     #
     # We'll just redo that quickly for usernames, to ensure they are safe.
 
-    user = re.sub(r'[^a-zA-Z0-9_.@-]', '_', user)
+    user = re.sub(rb'[^a-zA-Z0-9_.@-]', '_', user)
 
     def find_user(wanted):
-        with file('{{ openvpn_data_dir }}/users') as f:
+        with open('/etc/openvpn/cephco-sepia.data/users', 'rb') as f:
```
zmc, line 42: Why drop the templating here? Does the role need adjustment elsewhere to complement this change?

zmc (changes requested):

roles/gateway/templates/auth-openvpn
dmick: Mistake.  I was applying changes I'd actually tested by modifying the source, and didn't notice the change of template; was focused on the 'rb'.  fixing.

roles/gateway/templates/auth-openvpn
dmick: I swear I tested this, but I don't know how you're not right.  Looking.

dmick: actually, no, changing the split string to b' ' means that the outputs are also bytes.

roles/gateway/templates/auth-openvpn
dmick: forcepushed a fix for the dropped template.

roles/gateway/templates/auth-openvpn
zmc: The issue is with the line:
`salt = 'not-found'`
because:
```
>>> inner = hashlib.new('sha256')
>>> salt = 'not-found'
>>> inner.update(salt)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Strings must be encoded before hashing
```

zmc (changes requested):

roles/gateway/templates/auth-openvpn
dmick: Got that and the won't-match 'correct' string too.

zmc (approved):
🍻

PR #721, opened 2023-02-15
Add checkcerts, to use with cron to warn about expiring certs

dmick:
This originally lived on gitbuilder-archive, and I've moved it, revamped it, added some args, added some hosts, modified some emails, ported to Py3, stopped using external programs.  It's quick to run in default mode where it just reports to the terminal; it'll also be quiet and only send email about old certs.

The timezone processing is nonexistent on the reported expiry date; Python timezone handling is a mess.  That could be improved but not without a deep dive.

### DIFF
--- a/tools/checkcerts.py
+++ b/tools/checkcerts.py
@@ -1,0 +1,110 @@
+#!/usr/bin/python3
+
+import argparse
+import socket
+import ssl
+import subprocess
+import sys
+import os
+import tempfile
+import datetime
+import smtplib
+
+DAYS_BEFORE_WARN=7
+
+DEFAULT_DOMAINS = [
+    '1.chacra.ceph.com',
+    '2.chacra.ceph.com',
+    '3.chacra.ceph.com',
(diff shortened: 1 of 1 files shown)


dmick:
@zmc

dmick:
Added a couple of hosts, sorted the list

tools/checkcerts.py
```diff
@@ -0,0 +1,110 @@
...
+import socket
+import ssl
+import subprocess
+import sys
+import os
+import tempfile
+import datetime
+import smtplib
+
+DAYS_BEFORE_WARN=3
```
zmc, line 13: Should we make this a bit longer? We're in trouble if a cert wants to expire at the end of a long weekend.

dmick: I have no problem with that

dmick: Updated to 7 days

zmc (approved):

PR #722, opened 2023-02-27
paddles: Updates required for migration to new host

zmc:
python-docker doesn't exist on 20.04; python3-docker does, and also on 18.04, and requires python3-requests.


dmick (dismissed):
👍

dmick (approved):

PR #723, opened 2023-03-01
paddles: Deploy postgres-exporter

zmc:
When in containerized mode.


zmc:
Using docker-swarm here so that we can use a secret to store the password

dmick (approved):

PR #724, opened 2023-03-15
Add sentry.ceph.com to checkcerts.py

dmick:


zmc (approved):
Thanks for adding this!

For posterity, the sentry cert had expired today, and I'd gone and run `certbot renew`, wondering why we had let it expire. This will warn us next time.

PR #726, opened 2023-04-12
common: use EPEL metalink

ktdreyer:
Some mirrors are stale (https://pagure.io/fedora-infrastructure/issue/11233)

Use MirrorManager's metalink application so we always get up-to-date mirrors.

MirrorManager will also return the list of mirror that carry each architecture (x86_64, aarch64, etc) so we will not need to manage that information ourselves here.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -17,13 +17,13 @@ epel_mirror_baseurl: "http://dl.fedoraproject.org/pub/epel"
 epel_repos:
   epel:
     name: "Extra Packages for Enterprise Linux"
-    mirrorlist: file:///etc/yum.repos.d/epel-mirrorlist
+    metalink: "https://mirrors.fedoraproject.org/metalink?repo=epel-$releasever&arch=$basearch&infra=$infra&content=$contentdir"
     # ternary requires ansible >= 1.9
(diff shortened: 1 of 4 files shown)


ktdreyer:
This reverts the decision to use local mirrors, https://github.com/ceph/ceph-cm-ansible/pull/142

@dmick mentioned that we've had issues with Fedora's mirrorlist configuration in the past. I _think_ that removing `baseurl` (https://github.com/ceph/ceph-cm-ansible/pull/544) would make this more resilient

ktdreyer:
(also nice to stop using only plaintext `http://` URLs, especially since we have `gpgcheck: 0` at the moment)

ktdreyer:
Our static mirrorlist file has `download-ib01.fedoraproject.org` as the first item in the list. Ordinarily that server would be stable, since the Fedora admins use it as a primary mirror, but as https://pagure.io/fedora-infrastructure/issue/11233 shows, that `ib01` is now having trouble.

MirrorManager does not return this problematic mirror:

```
curl 'https://mirrors.fedoraproject.org/metalink?repo=epel-8&arch=x86_64' | grep ib01
```

ljflores (approved):

akraitman (approved):

PR #727, opened 2023-04-13
Removing broken epel mirrors and adding the one that's working

akraitman:
Error - Status code: 404 for http://mirror.pnl.gov/epel/8/Everything/x86_64/repodata/be7aa5c233d3e92cbff3245ce5a5ded68256363f9b2d0f18be6df7caf9b0c674-filelists.xml.gz

### DIFF
--- a/roles/common/templates/epel-mirrorlist
+++ b/roles/common/templates/epel-mirrorlist
@@ -1,9 +1,4 @@
 # {{ ansible_managed }}
 
 # local yum mirrorlist for epel-{{ ansible_distribution_major_version }}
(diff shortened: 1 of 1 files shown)


akraitman:
Fedora infra team just wrote "I am still trying to get download-ib01 back fully in sync. It's better now, but still not 100% caught up yet.

We notified the kernel.org folks. They were syncing from download-ib01. We asked them to move and sync from dl.fedoraproject.org at least for now, so hopefully they will be back in sync again soon."

ktdreyer:
Closed for https://github.com/ceph/ceph-cm-ansible/pull/726

PR #728, opened 2023-04-13
Adding dl.fedoraproject.org mirror temporality until download-ib01 wi…

akraitman:
…ll be back


ktdreyer:
I would prefer we do https://github.com/ceph/ceph-cm-ansible/pull/726

ljflores (approved):

PR #730, opened 2023-04-17
common: remove RHEL beta handling

ktdreyer:
Now that mainly care about containers and we test against CentOS Stream, there's almost no value in testing Ceph with RHEL Beta content.

The `rhel-entitlements.yml` playbook is very complicated, and dropping this feature will make it easier to simply and refactor further.


ktdreyer:
I'm removing code here that I found to be buggy in some corner-cases and required further maintenance:
- The playbook never succeeds if `have_entitlements` is `false` (`subscription-manager unregister` always terminates with a non-zero exit status when a node is not registered)
- `ansible_lsb.description` no longer exists, so `Beta` will never be found.
- `beta_repos.yml` ought to use [`yum_repository`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_repository_module.html) instead of our custom `yum_repo.j2` template

akraitman (approved):

PR #731, opened 2023-04-20
Remove local mirrorlist cache for Centos8 testhosts

dmick:


ktdreyer:
You could delete this bit too, from the top of `roles/testnode/tasks/yum/repos.yml`

```yaml
- name: Configure local mirrorlists
  template:
    src: 'mirrorlists/{{ ansible_distribution_major_version }}/{{ item }}'
    dest: '/etc/yum.repos.d/{{ item }}'
    owner: root
    group: root
    mode: 0644
  with_items: "{{ yum_mirrorlists }}"
  when: yum_mirrorlists is defined
```

ktdreyer:
And some references in `roles/testnode/README.rst` ... I think `git grep mirrorlist`

dmick:
Yeah, I left those other things because they're still capabilities if we ever need to have them back, but should be harmless

dmick:
I've tested with a local centos8stream kvm; the common.yml playbook runs to completion; the Centos8 repos are still using mirrorlist=http://mirrorlist.centos.org... and dnf update works fine.  epel is installed and good.  I think this is ready for deployment.

ktdreyer:
I created a CentOS 8 Stream VM for testing, and didn't see anything changing... then I realized this only applies to CentOS 8, not CentOS 8 Stream.

We should convert everything in the lab to from CentOS 8 to CentOS 8 Stream, because CentOS 8 has many security flaws and the developers haven't pushed updates since December 31st, 2021

ktdreyer:
Ok, I discovered more information while testing this.

## Testing with the correct playbook
I needed to run the `testnodes.yml` playbook for my VM (not the `common.yml` playbook.)

## Repo names

`roles/testnode/vars/centos_8.yml` used the old `/etc/yum.repos.d/CentOS-*.repo` filenames. The CentOS project developers have changed those default filenames over time, first to add `-Linux`, and finally to have `-Stream`:

| [cloud image](https://cloud.centos.org/centos/8/x86_64/images/) tested | Default `/etc/yum.repos.d`  | Ships in pkg                      |
|------------------------------------------------------------------------|-----------------------------|-----------------------------------|
| `CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2`               | `CentOS-BaseOS.repo`        | `centos-repos-8.2-2.2004.0.1.el8` |
| `CentOS-8-GenericCloud-8.3.2011-20201204.2.x86_64.qcow2`               | `CentOS-Linux-BaseOS.repo`  | `centos-linux-repos-8-2.el8`      |
| `CentOS-8-GenericCloud-8.4.2105-20210603.0.x86_64.qcow2`               | `CentOS-Linux-BaseOS.repo`  | `centos-linux-repos-8-2.el8`      |
| CentOS 8 Stream                                                        | `CentOS-Stream-BaseOS.repo` | `centos-stream-repos-8-6.el8`     |

`centos_8.yml` had a comment "These will overwrite the repo files that come with a CentOS installation", and apparently this stopped being true with CentOS 8.3.

## Stream vs Non-Stream

With this PR, we stop writing the old `/etc/yum.repos.d/CentOS-BaseOS.repo` entirely (good). (To delete the file from existing nodes, though, we'll need to reimage.)

*For non-Stream nodes:* if they have the "newer" repository definitions (`/etc/yum.repos.d/CentOS-Linux-BaseOS.repo`) from CentOS 8.3+, then those will probably use CentOS' MirrorManager. That's a problem because CentOS 8 "non-stream" is EOL, and MirrorManager no longer serves non-Stream content. We should plan to reimage any non-Stream nodes to Stream, though.

*For Stream nodes:* David realized the repo names were different with CentOS 8 Stream, so he wrote `roles/testnode/tasks/yum/stream.yml` that runs `rm -rvf /etc/yum.repos.d/*-Stream-*`. Importantly, with this PR, Stream nodes end up with *no* CentOS repositories in `/etc/yum.repos.d`. **We should delete this `stream.yml` as part of this PR.**

dmick:
whew.  a few comments:  

1. testnodes get reimaged before each run, so the effects should be quick for testnodes.  There may be things that need fixing on jenkins build nodes too.
2. 'original' repos should be able to be restored by installing the centos-stream-repos package (for Stream, obviously) and I assume there are similar for other releases if we're not just going to get rid of them

I'm gonna go look at reimage stats and see if we're using any centos8 besides stream.  I don't see why we should at this point.

ktdreyer:
When I looked at https://pulpito.ceph.com/nodes/ and sorted by OS Version, I saw `8`, `8.4`, `8.6`, and `8.stream`, and I wasn't sure what that implied for this PR.

I do think we should delete `roles/testnode/tasks/yum/stream.yml` as part of this PR.

dmick:
Yeah, I'm testing that now

dmick:
OK, with --skip-tags=lvm, it completes just fine.  (Guess what OS installs to lvm by default.  Guess what the playbook destroys. sigh.)  I had a libvirt vm installed from CentOS-Stream-8-x86_64-latest-boot.iso, and ran the playbook against it.  Afterward, I ran "dnf makecache" with no errors.  I also did that after dnf config-manager --enable powertools, and it succeeded in making the cache for powertools as well.  I think this is ready to merge.

akraitman (approved):

PR #734, opened 2023-04-24
Ansible Role for deploying grafana-agent

akraitman:


roles/grafana_agent/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: "Restart grafana agent instance - linux"
```
ktdreyer, line 2: Simpler is better
```suggestion
- name: "Restart grafana agent"
```

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+
+# Deploy config file from template and restart the agent      
+- name: "Configure agent"
+  become: true
+  ansible.builtin.template:
+    src: "templates/linux_agent_config.yml.j2"
+    dest: "/etc/grafana-agent.yaml"
+    mode: "0550"
+    group: "grafana-agent"
+  notify: "Restart grafana agent instance - linux"
```
ktdreyer, line 51:
```suggestion
  notify: "Restart grafana agent"
```

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+- name: "Add Grafana's repository to APT sources list"
+  become: true
+  ansible.builtin.apt_repository:
+    repo: "deb https://apt.grafana.com stable main"
+    state: present
+  when: ansible_pkg_mgr == "apt"
+
+- name: "Add Grafana's repository to yum/dnf systems"
+  become: true
+  ansible.builtin.template:
```
ktdreyer, line 24: Several ceph-cm-ansible roles were written in 2015 before Ansible had a [`yum_repository` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_repository_module.html), and I understand this was probably copied from that pattern.

 It's much better to use the `yum_repository` module than to write our own `.repo` file with `template`.

(One example of the complexity that goes away with `yum_module`: the `mode` parameter should not be `0655` for a `.repo` file. That would make the file executable.)

roles/grafana_agent/templates/grafana_repo.yml.j2
```diff
@@ -0,0 +1,9 @@
+[grafana]
+name=grafana
+baseurl=https://rpm.grafana.com
+repo_gpgcheck=1
+enabled=1
+gpgcheck=1
+gpgkey=https://rpm.grafana.com/gpg.key
+sslverify=1
+sslcacert=/etc/pki/tls/certs/ca-bundle.crt
```
ktdreyer, line 9: The `sslverify` and `sslcacert` parameters are not necessary. Don't bother copying them when you rewrite this to use `yum_module`.

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+    state: "started"
+    enabled: true
+
+# Deploy config file from template and restart the agent      
+- name: "Configure agent"
+  become: true
+  ansible.builtin.template:
+    src: "templates/linux_agent_config.yml.j2"
+    dest: "/etc/grafana-agent.yaml"
+    mode: "0550"
```
ktdreyer, line 49: I recommend renaming the template to `grafana-agent.yaml.j2` so that it matches what you write to /etc. When those file names do not match, it makes it more confusing to maintain this codebase.

We should have a `user` parameter here.

`mode: "0550"` does not look right here. This file should not be executable. Probably `mode: "0440"` would be right, assuming `user` will be `root`.

ktdreyer (changes requested):
This is a good start and there are some things we could improve.

I recommend merging https://github.com/ceph/ceph-cm-ansible/pull/733 first, then rebasing this PR onto `main`, so that the linter will run on this PR.

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+    repo: "deb https://apt.grafana.com stable main"
+    state: present
+  when: ansible_pkg_mgr == "apt"
+
+- name: "Add Grafana's repository to yum/dnf systems"
+  become: true
+  ansible.builtin.yum_repository:
+    baseurl: "https://rpm.grafana.com"
+    name: "grafana"
+    description: "Grafana's repository"
```
ktdreyer, line 27: Please enable GPG checking:
```suggestion
    description: "grafana"
    gpgcheck: true
    gpgkey: https://rpm.grafana.com/gpg.key
```

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
+---
+- name: Include secrets
+  include_vars: "{{ secrets_path | mandatory }}/mimir_password.yml"
+  no_log: true
+  tags:
+    - always
+
+- name: "Import Grafana GPG key"
+  become: true
+  ansible.builtin.apt_key:
```
ktdreyer, line 10: I understand we've used `apt_key` for a while, but as [the `apt_key` docs explain](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html), the latest Debian/Ubuntu will no longer support this.

It's ok for today, but in the future, prepare to use the alternative example in the [`apt_repository` docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_repository_module.html) (the `get_url` module plus the `signed-by` setting).

ktdreyer (dismissed):

roles/grafana_agent/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
...
+
+# Mimir URL and creds
+agent_mimir_url: "http://sepia-grafana.front.sepia.ceph.com:9009/api/v1/push"
+agent_mimir_username: "admin"
+
+# Enable metrics globally
+agent_metrics: true
+
+# Labels used for external labels
+nodetype: "unknown_nodetype"
```
ktdreyer, line 12: Why is this a variable? Can we hard-code it into `grafana-agent.yaml.j2` for simplicity?

roles/grafana_agent/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+# defaults file for agent
+
+# Mimir URL and creds
+agent_mimir_url: "http://sepia-grafana.front.sepia.ceph.com:9009/api/v1/push"
+agent_mimir_username: "admin"
+
+# Enable metrics globally
+agent_metrics: true
```
ktdreyer, line 9: Why does this conditional exist? Do you expect that certain hosts will set this to `false` in a future PR?

roles/grafana_agent/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+# defaults file for agent
```
ktdreyer, line 2: This comment is superfluous. Would you please remove it?

roles/grafana_agent/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
...
+agent_mimir_url: "http://sepia-grafana.front.sepia.ceph.com:9009/api/v1/push"
+agent_mimir_username: "admin"
+
+# Enable metrics globally
+agent_metrics: true
+
+# Labels used for external labels
+nodetype: "unknown_nodetype"
+
+scrape_interval_global: "60s"
```
ktdreyer, line 14: Why not put these values directly into `grafana-agent.yaml.j2`?

We should only template things that will change host-to-host, or things we expect other (non-Sepia) lab administrators to alter.

roles/grafana_agent/defaults/main.yml
akraitman: Those labels are used in Prometheus to distinguish between the different machines groups

roles/grafana_agent/defaults/main.yml
akraitman: This configuration is important because it determines the number of data points per minute (DPM) scraped in your active series. Depending on how you ship metrics to Grafana from the agent, you can reduce your total DPM by adjusting the scrape_interval setting, the same for the "scrape_interval_node" We can change those in the future in case overpopulated the Grafana server

roles/grafana_agent/defaults/main.yml
ktdreyer: I understand. My point is that in your PR, it's always `true`. Why not simply remove it?

ktdreyer:
Could you merge https://github.com/ceph/ceph-cm-ansible/pull/733 first, then rebase this PR onto `main`? Then the linter will run on this PR.

ktdreyer:
Adam and I chatted about this PR.

* These variables exist because Adam sets them with `extra_vars` when running the playbook interactively on his computer.
* We discussed the possibility of tracking those particular non-default choices in `host_vars` or `group_vars` instead so that we can understand how and why Adam made those choices to alter the values from the defaults.

dmick:
@akraitman have you resolved all of Ken's earlier issues?

akraitman:
> @akraitman have you resolved all of Ken's earlier issues?

Yes, I removed the if statement  grafana-agent.yaml.j2 and removed the "agent_metrics" variable from defaults/main.yml

roles/grafana_agent/defaults/main.yml
akraitman: I Placed nodetype directly in grafana-agent.yaml.j2

roles/grafana_agent/tasks/main.yml
dmick: We should do this the new way, and we should also work on fixing all the rest of the apt-key invocations.  One of the Debs already requires it and I'm not sure but I think newer Ubuntu may as well.

akraitman: > We should do this the new way, and we should also work on fixing all the rest of the apt-key invocations. One of the Debs already requires it and I'm not sure but I think newer Ubuntu may as well.

I updated the key import task to ansible.builtin.get_url and tested it on apt/rpm machines.

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+  ansible.builtin.package:
+    name: "grafana-agent"
+    state: "present"
+
+- name: "Making sure prometheus service is not running since its using the same port as grafana-agent 9090"
+  become: true
+  ansible.builtin.service:
+    name: "prometheus"
+    state: "stopped"
+    enabled: false
```
djgalloway, line 50: This is kind of dangerous.  Are we running prometheus anywhere?

djgalloway: And won't the next task just fail if we are anyway?

akraitman: Yes we are currently running prometheus only on the sepia-grafana server, I think the change I pushed right now is a better solution, the playbook will only check if prometheus listens and it won't continue until it's resolved

djgalloway (dismissed):

roles/grafana_agent/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+- name: Include secrets
+  include_vars: "{{ secrets_path | mandatory }}/mimir_password.yml"
+  no_log: true
+  tags:
+    - always
+
+- name: Gather facts on listening ports
+  community.general.listen_ports_facts:
+
+- name: Check if prometheus is lisening on port 9090
```
dmick, line 11: listening

djgalloway (approved):

PR #740, opened 2023-05-16
common/tasks/rhel-entitlements.yml: regenerate machine-id

dmick:
We can't leave /etc/machine-id blank; it breaks things, one of which is the kernel install, which runs a postinstall script to update /boot/loader/entries, which does nothing (silently) if there's nothing in /etc/machine-id.  Since it can come from the dbus id, and does by default, and there's a command to generate the dbus id, generate both, dbus first.  This fixes the kernel postinstall. I don't know if there should be any correlation between machine-id and the subscription-manager/katello IDs.

### DIFF
--- a/roles/common/tasks/rhel-entitlements.yml
+++ b/roles/common/tasks/rhel-entitlements.yml
@@ -50,6 +50,12 @@
   register: new_uuid
   when: use_satellite == true
 
+- name: Run dbus-uuidgen to create /var/lib/dbus/machine-id
+  shell: dbus-uuidgen --ensure
+
+- name: Run systemd-machine-id-setup to set /etc/machine-id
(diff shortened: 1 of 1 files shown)


zmc (approved):

roles/common/tasks/rhel-entitlements.yml
akraitman, line 1: Hey @dmick I think we should run those tasks after the "subscription-manager register" task, the reason why we remove those files in the previous task is to workaround the issue we had when we moved to satellite 6.12 where machines wore unable to subscribe - https://access.redhat.com/support/cases/#/case/03399123

dmick: agreed

zmc (approved):

PR #742, opened 2023-06-02
testnode: Make sure PowerTools repo is enabled on CentOS

dmick:
https://github.com/ceph/ceph-cm-ansible/pull/731 removed the custom-made repo files that added mirrorlists; however, it also removed the side-effect of enabling the Power Tools repo (which is not enabled by default).  This adds a call to dnf config-manager to enable the repo, whatever its repo file name, on CentOS testnodes.

Fixes: https://tracker.ceph.com/issues/59678

### DIFF
--- a/roles/testnode/tasks/yum/repos.yml
+++ b/roles/testnode/tasks/yum/repos.yml
@@ -38,6 +38,11 @@
     - (ansible_os_family == "RedHat" and ansible_distribution_major_version|int >= 8)
     - copr_repos|length > 0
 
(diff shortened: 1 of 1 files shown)


dmick:
I tested this by manually running the testnodes playbook against a CentOS 8 Stream host I'd provisioned with teuthology-lock

zmc (approved):

cbodley:
i opened https://tracker.ceph.com/issues/61648 to track this. i set its priority to urgent, because centos9 support will be a blocker for the reef release

PR #746, opened 2023-08-02
centos8, centos9: install dbench package from epel instead of lab-extras

lxbsz:
Fixes: https://tracker.ceph.com/issues/62227

### DIFF
--- a/roles/testnode/vars/centos_8.yml
+++ b/roles/testnode/vars/centos_8.yml
@@ -61,13 +61,13 @@ packages:
   # for pjd tests
   - libacl-devel
   # for fs tests,
-  - dbench
   - autoconf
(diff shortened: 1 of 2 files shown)


roles/testnode/tasks/yum/packages.yml
```diff
@@ -57,6 +57,13 @@
     enablerepo: epel
   when: epel_packages|length > 0
 
+- name: Install copr packages
```
dmick, line 60: why is this mentioning "copr"?  if it's epel?

lxbsz: Ah, sorry, I forgot to remove this.

dmick (approved):

PR #747, opened 2023-08-02
centos_8_stream: install dbench package from epel

lxbsz:
Fixes: https://tracker.ceph.com/issues/62227


dmick:
As in https://github.com/ceph/ceph-cm-ansible/pull/746

dmick (approved):

PR #748, opened 2023-08-25
roles/testnode/vars/centos_9.yml: add lab-extras and iozone

dmick:
iozone is present in the newly-created lab-extras repo, for the same reasons it's in other CentOS configs

### DIFF
--- a/roles/testnode/vars/centos_9.yml
+++ b/roles/testnode/vars/centos_9.yml
@@ -1,7 +1,12 @@
 ---
 # vars specific to any centos 9.x version
 
-common_yum_repos: {}
+common_yum_repos:
+  lab-extras:
+    name: "lab-extras"
+    baseurl: "http://{{ mirror_host }}/lab-extras/9/"
+    enabled: 1
+    gpgcheck: 0
 
 yum_repos:
   CentOS-AppStream:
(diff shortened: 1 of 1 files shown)


dmick:
this needs to be tested before merging

idryomov (approved):
Looks "obviously correct" to me.

idryomov:
> this needs to be tested before merging

Just curious, is there now a way pick a custom ceph-cm-ansible branch into a teuthology-suite run?  Or are you referring to the https://github.com/ceph/ceph-cm-ansible#setting-up-a-local-dev-environment procedure?

dmick:
@idryomov the reason I said "needs to be tested" is because I'm not sure :)  Honestly I'm not even certain which roles trigger this code.  I'll find out.

dmick:
As for running a branch of ceph-cm-ansible during teuthology, yes, there's a mechanism for that, see https://github.com/ceph/teuthology/blob/main/teuthology/task/ansible.py#L468C1-L485C34

dmick:
That said, probably a sufficient test here is just to lock a centos9 machine and then run testnodes after the fact

dmick:
Did the test proposed above, host was left with active lab-extras repo and an installed iozone package

PR #749, opened 2023-11-29
roles/common: Enable codeready repos on RHEL

zmc:


zmc:
dnf says `Error: 'codeready-builder-for-rhel-8-x86_64-rpms' does not match a valid repository ID.`

dmick (approved):
I think this would be the right thing to do if we could convince our Satellite instance to serve that repo.  I'm trying to make that happen; we should hold off merging this until that gets sorted, or else every rhel provision may break (it seems like maybe the --enable works, but then later dnf operations fail?  I'm not sure)  Anyway, the code change is sound, but our infra isn't ready for it yet

dmick:
Satellite is fixed up now.  This should work, and is necessary

PR #750, opened 2023-12-12
rhel-entitlements.yml: Avoid releasever hack on all but el7

dmick:


dmick:
Tested with a local run of testnode.yml, seems to do the right thing on 8.6 at least.

zmc (approved):

PR #751, opened 2023-12-13
cobbler_profile: fix two bugs discovered while importing inktank-rescue

dmick:
1) ansible's get_url moves from sha256 to checksum 2) fix 'registered' name (import -> imported) in import_distro_image


dmick:
@akraitman fyi

akraitman:
> @akraitman fyi

Thank you @dmick

zmc (approved):
Looks good; verified the syntax with the documentation as well.

PR #753, opened 2024-03-06
testnode: Fix CentOS 9 BaseOS/AppStream repos

zmc:
/test/ now points to 8.Stream, for whatever reason.

### DIFF
--- a/roles/testnode/vars/centos_9.yml
+++ b/roles/testnode/vars/centos_9.yml
@@ -11,13 +11,13 @@ common_yum_repos:
 yum_repos:
   CentOS-AppStream:
     name: "CentOS-$releasever - AppStream"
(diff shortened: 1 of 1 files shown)


dmick:
I don't think we should be using those URLs, but have not yet investigated what the right answer is

dmick:
So this is the default AppStream selection for stream8:
mirrorlist=http://mirrorlist.centos.org/?release=$stream&arch=$basearch&repo=AppStream&infra=$infra
#baseurl=http://mirror.centos.org/$contentdir/$stream/AppStream/$basearch/os/

and stream9:

metalink=https://mirrors.centos.org/metalink?repo=centos-appstream-$stream&arch=$basearch&protocol=https,http

Those are so far from what the ansible is doing that I'm afraid we're in horribly-unsupported land and they're apparently going back and forth.  I don't know how to infer their intent with "test" and "development" and I'm not convinced I can make the right choice that will also be permanent.

PR #760, opened 2024-09-30
testnode: remove tgt

idryomov:
As a follow up for commit 67a92953a5a2 ("testnode: don't install tgt"), explicitly remove tgt to cover the case of it already being there.  This also documents that we actively don't want tgt to be there.


idryomov:
This is a follow up for #759.

akraitman (approved):

PR #761, opened 2024-10-01
testnode: move removing openmpi-common to apt_systems.yml

idryomov:
Otherwise, because it's placed in ubuntu.yml, it resets packages_to_remove list defined in apt_systems.yml to just openmpi-common and commit 701d3594d220 ("testnode: remove tgt") doesn't take effect.

While at it, fix a typo in the comment -- it's mpich.


idryomov:
This is a follow up for https://github.com/ceph/ceph-cm-ansible/pull/760.

akraitman (approved):

PR #764, opened 2025-01-30
Modyfing openVPN logrotate configuration to keep the logs for a year.

falcocer-ibm:
Fixes: https://ibm.monday.com/boards/5591222586/pulses/8354436354

### DIFF
--- a/roles/gateway/files/openvpn.logrotate
+++ b/roles/gateway/files/openvpn.logrotate
@@ -1,6 +1,6 @@
 /var/log/openvpn/*.log {
 	daily
-	rotate 90
+	rotate 366
 	compress
 	missingok
 	copytruncate


djgalloway:
Oh, that was easier than I expected 😄.  I'm just taking a look at the log file sizes on the machine and think we should probably change them to rotate weekly as well.  Can you make that change too please?

```
[root@gw openvpn]# ls -lah
total 313M
drwxr-xr-x.  2 root root  4.0K Jan 30 03:37 .
drwxr-xr-x. 17 root root  4.0K Jan 26 03:32 ..
-rw-------   1 root root   90M Jan 30 18:07 openvpn.log
-rw-------   1 root root  258K Nov  2 03:06 openvpn.log-20241102.gz
-rw-------   1 root root  231K Nov  3 03:18 openvpn.log-20241103.gz
-rw-------   1 root root  241K Nov  4 03:44 openvpn.log-20241104.gz
-rw-------   1 root root  296K Nov  5 03:20 openvpn.log-20241105.gz
-rw-------   1 root root  286K Nov  6 03:30 openvpn.log-20241106.gz
-rw-------   1 root root  664K Nov  7 03:33 openvpn.log-20241107.gz
-rw-------   1 root root  707K Nov  8 03:39 openvpn.log-20241108.gz
-rw-------   1 root root  375K Nov  9 03:07 openvpn.log-20241109.gz
<snip>
```

falcocer-ibm:
I thought on that too, and yes based on the partition size I think that could works, but I rather to keep it daily because is easier to filter logs by day. But sure, I will make the change.

djgalloway:
Maybe if there's a way to keep, say, a month of dailies and then 51 weeklies?  IDK, anything seems cleaner than 366 daily log files IMO.

PR #767, opened 2025-02-18
checkcerts.py: actually fix "send email"

dmick:
argparse can't do a nargs="*" optional arg *and* check for its presence; add a separate arg -E to send the email, and keep -e as an optional list of addressees.

Also add the full path and host where checkcerts.py is running.


djgalloway (approved):

djgalloway:
Tested? Working?

dmick:
yes, to ceph-infra@redhat.com and selected individuals (the default in the python).  Perhaps we should have a less-rh-centric team alias instead, but I got emails about [14].chacra.ceph.com this week

PR #770, opened 2025-02-28
tools/checkcerts.py: fix deprecated datetime.datetime.utcnow()

dmick:
The replacement, datetime.datetime.now(datetime.UTC), returns a
    tz-aware item, so the expiry time must also be timezone-aware.
    datetime.strptime() does not handle that time format sanely.
    I don't know why.

    Luckily zoneinfo exists and can parse likely timezones into valid
    datetime.tzinfo instances, as far as I can tell.  Why isn't this part
    of datetime?  I don't know.  But this code seems be somewhat reliable
    (assuming the timezone name is the last item in the date string,
    delimited by spaces).

### DIFF
--- a/tools/checkcerts.py
+++ b/tools/checkcerts.py
@@ -9,6 +9,7 @@ import os
 import tempfile
 import datetime
 import smtplib
+import zoneinfo
 
 DAYS_BEFORE_WARN=7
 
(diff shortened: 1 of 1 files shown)


dmick:
replaced original change with a better idea using the new "zoneinfo".  It will require Python 3.9, which is not currently on gw, so I'll find a better place to run this.

dmick:
eh.  this is required by and requires a pretty-recent python version (later even than 3.9).  In retrospect I think I'll hold off on this until we decide if we even want to continue with checkcerts.py as we update our infra.  Adam already has some grafana alerts set up for them and that may be a better plan, particularly if we can change the alerting scheme from the Slack-only verbose format it has now.

djgalloway:
This is handled by cert-manager and alerts

PR #771, opened 2025-02-28
workflows: Install and use ansible in a virtualenv

djgalloway:


djgalloway:
`git commit --amend --edit` and add `workflows:` before the commit message and I'm happy 😄 .  Fantastic work finding this.

.ansible-lint
```diff
@@ -11,7 +11,7 @@ skip_list:
   - git-latest
   - jinja
   - literal-compare
-  - load-failure
+#  - load-failure
```
zmc, line 14: Why comment as opposed to remove?

zmc (commented):
I wouldn't block this PR on this, but it might be worth considering using [pipx](https://pipx.pypa.io/) instead of manually managing a virtualenv. For comparison:
```
export DEBIAN_FRONTEND=noninteractive
apt-get update && apt-get install -y pipx
pipx ensurepath  # this sets PATH in ~/.profile and ~/.bashrc
export PATH=$PATH:$HOME/.local/bin  # this sets PATH in the current shell
pipx install --include-deps ansible-lint  # now ansible-* are all available to use

```

.github/workflows/tests.yml
```diff
@@ -11,28 +11,57 @@ jobs:
...
         ansible-playbook -i localhost, cephlab.yml --syntax-check
+
   ansible-lint:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Install ansible-lint
+  
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
```
djgalloway, line 33: Just pin this to a specific sha1 and we're good to go.

PR #773, opened 2025-03-03
testnode: Run grafana_agent role optionally

djgalloway:
Snippet from ansible.log
```
TASK [grafana_agent : Gather facts on listening ports] *************************
skipping: [smithi116.front.sepia.ceph.com] => {"changed": false, "false_condition": "run_grafana_agent_role is defined and run_grafana_agent_role", "skip_reason": "Conditional result was False"}
```

https://pulpito.ceph.com/dgalloway-2025-03-03_22:29:15-smoke-squid-distro-default-smithi/

We'll define this at the machine type level in the secrets repos.

### DIFF
--- a/roles/testnode/README.rst
+++ b/roles/testnode/README.rst
@@ -303,6 +303,11 @@ Configure ``cachefilesd``.  See https://tracker.ceph.com/issues/6373.  Defaults 
     cachefilesd_fstop
     cachefilesd_secctx
 
+Include the grafana_agent role to report testnode resource statistics to Grafana.  Defaults to ``false``::
+
+    run_grafana_agent_role: true
+
+
 Tags
 ++++
 
--- a/roles/testnode/meta/main.yml
+++ b/roles/testnode/meta/main.yml
@@ -1,3 +1,5 @@
 ---
 dependencies:
   - role: secrets
+  - role: grafana_agent
+    when: run_grafana_agent_role is defined and run_grafana_agent_role
--- a/testnodes.yml
+++ b/testnodes.yml
@@ -4,5 +4,4 @@
   roles:
     - common
     - testnode
-    - grafana_agent
   become: true


kshtsk:
Are those ansible-lint checks related to the PR? Seems not.

djgalloway:
> Are those ansible-lint checks related to the PR? Seems not.

No, those checks have been broken for a long time.  They're being worked on.

kshtsk:
What else do we need to let this PR merged?

zmc:
Just curious, can this not be accomplished with tags? I ask because we already use the `--skip-tags` feature to skip a number of tags, and this would prevent having to implement a teuthology feature just for this.

djgalloway:
> Just curious, can this not be accomplished with tags? I ask because we already use the `--skip-tags` feature to skip a number of tags, and this would prevent having to implement a teuthology feature just for this.

I considered that but landed on this approach instead to avoid adding `tags: grafana_agent` to each task in the role's [main.yml](https://github.com/ceph/ceph-cm-ansible/blob/main/roles/grafana_agent/tasks/main.yml)

kshtsk:
> > Just curious, can this not be accomplished with tags? I ask because we already use the `--skip-tags` feature to skip a number of tags, and this would prevent having to implement a teuthology feature just for this.
> 
> I considered that but landed on this approach instead to avoid adding `tags: grafana_agent` to each task in the role's [main.yml](https://github.com/ceph/ceph-cm-ansible/blob/main/roles/grafana_agent/tasks/main.yml)

Won't this work?

```
---
- hosts: testnodes
  strategy: free
   roles:
     - role: common
     - role: testnode
     - role: grafana_agent
        tags: graphana
   become: true
   ```

djgalloway:
> > > Just curious, can this not be accomplished with tags? I ask because we already use the `--skip-tags` feature to skip a number of tags, and this would prevent having to implement a teuthology feature just for this.
> > 
> > 
> > I considered that but landed on this approach instead to avoid adding `tags: grafana_agent` to each task in the role's [main.yml](https://github.com/ceph/ceph-cm-ansible/blob/main/roles/grafana_agent/tasks/main.yml)
> 
> Won't this work?
> 
> ```
> ---
> - hosts: testnodes
>   strategy: free
>    roles:
>      - role: common
>      - role: testnode
>      - role: grafana_agent
>        tags: graphana
>    become: true
> ```

I guess?  It still runs by default unless you `--skip-tags grafana` which, in hindsight, seems like not the right thing to do.

kshtsk (approved):

djgalloway:
No, I know what you were suggesting.  However, running the grafana_agent role on testnodes turned out to be a breaking change so it shouldn't have been enabled by default anyway is what I was trying to convey.

akraitman (approved):

akraitman:
Tested on one of the smithi machines and it looks good, skipping the grafana_agent role on testnodes by default

zmc:
CI is passing again! Thanks @djgalloway

kshtsk:
@djgalloway thank you

PR #775, opened 2025-03-20
tools: Exclude `option host-name` from grep hostname in set-next-server

djgalloway:
Otherwise, the `linenum=` line gets two line numbers.


tools/set-next-server.sh
```diff
@@ -37,7 +37,7 @@ fogip="172.21.0.72"
 fogfilename="/undionly.kpxe"
 macaddr=$(sed -n "/host ${host}-front/,/}/p" $dhcpconfig | grep 'hardware ethernet' | awk '{ print $3 }' | tr -d ';')
 ipaddr=$(sed -n "/host ${host}-front/,/}/p" $dhcpconfig | grep 'fixed-address' | awk '{ print $2 }' | tr -d ';')
-linenum=$(grep -n $host $dhcpconfig | cut -d ':' -f1)
+linenum=$(grep -n $host $dhcpconfig | grep -v "host-name" | cut -d ':' -f1)
```
dmick, line 41: I probably would have changed the grep to "$host $host-front {" to match the sed, but I guess this works up until we add some other option with the hostname in it

dmick (approved):

PR #778, opened 2025-03-25
maas dns role for configuring maas dns domains and records

akraitman:


djgalloway:
Please add a README explaining all of the variables and what the secrets variables should look like.  Like this: https://github.com/ceph/ceph-cm-ansible/blob/main/roles/nameserver/README.rst

roles/maas_nameserver/README.rst
```diff
@@ -0,0 +1,102 @@
...
+   ansible-playbook maas-dns-playbook.yml
+   ```
+
+## Variables
+
+### `defaults/main.yml`
+These are overridable defaults:
+- `maas_api_url`: Default MAAS API endpoint (`http://localhost:5240/MAAS/api/2.0/`). Typically overridden by the secrets file.
+- `maas_profile`: Default MAAS profile name (`admin`). Overridden by the secrets file.
+- `allowed_domains`: List of domains to preserve (default: `["maas"]`).
```
djgalloway, line 55: What does "to preserve" mean?

roles/maas_nameserver/README.rst
```diff
@@ -0,0 +1,102 @@
...
+
+### `vars/main.yml`
+These are role-specific variables, not intended for override:
+- `dns_domains`:
+  - `ceph`: Dynamically derived from the first Ceph host’s domain (e.g., `internal.ceph.ibm.com`).
+  - `ipmi`: Static IPMI domain (`ipmi.ceph.ibm.com`).
+  
+## Secrets File
+The `maas_nameserver` role depends on the `secrets` role to load encrypted MAAS API credentials. The secrets file is expected at `{{ secrets_path }}/group_vars/maas.yml`, where `secrets_path` defaults to `/etc/ansible/secrets` (configurable via the `ANSIBLE_SECRETS_PATH` environment variable).
```
djgalloway, line 64: That's not what [this](https://github.com/ceph/ceph-cm-ansible/blob/5fed40f28bd2976d60cd526a8e812482e868a02c/roles/secrets/tasks/main.yml#L4) says.

djgalloway:
If you look at the README in the [Web UI](https://github.com/ceph/ceph-cm-ansible/blob/5fed40f28bd2976d60cd526a8e812482e868a02c/roles/maas_nameserver/README.rst), it does not get rendered as expected.  Maybe it's `.md` instead of .rst`?

djgalloway (changes requested):

roles/maas_nameserver/README.rst
akraitman: There is a default domain the maas server created called "maas" and I am not sure it can be removed so I added this "allowed_domains" variable to ignore this domain, so preserve means not to delete it

roles/maas_nameserver/tasks/main.yml
```diff
@@ -0,0 +1,95 @@
...
+    loop_var: host
+  when: "'ipmi' in hostvars[host]"
+
+- name: Parse desired FQDNs
+  ansible.builtin.set_fact:
+    desired_fqdns: "{{ dns_records | map(attribute='name') | zip(dns_records | map(attribute='domain')) | map('join', '.') | list }}"
+
+- name: Debug desired FQDNs
+  ansible.builtin.debug:
+    msg: "Desired FQDNs: {{ desired_fqdns }}"
```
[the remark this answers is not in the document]
akraitman: TASK [maas_nameserver : Debug desired FQDNs] **************************************************************************************************************************************************
ok: [ceph-vm-14.tuc.stglabs.ibm.com] => {
    "msg": "Desired FQDNs: ['ceph001.internal.ceph.ibm.com', 'ceph002.internal.ceph.ibm.com', 'ceph003.internal.ceph.ibm.com', 'ceph001.ipmi.ceph.ibm.com', 'ceph002.ipmi.ceph.ibm.com', 'ceph003.ipmi.ceph.ibm.com']"
}

roles/maas_nameserver/README.rst
akraitman: I fixed this line in the readme file

roles/maas_nameserver/tasks/main.yml
```diff
@@ -0,0 +1,95 @@
...
+  loop_control:
+    loop_var: host
+
+- name: Build DNS records for IPMI interfaces
+  ansible.builtin.set_fact:
+    dns_records: "{{ dns_records + [{'name': host.split('.')[0], 'ip': hostvars[host]['ipmi'], 'type': 'A', 'domain': dns_domains.ipmi}] }}"
+  loop: "{{ groups['ceph'] }}"
+  loop_control:
+    loop_var: host
+  when: "'ipmi' in hostvars[host]"
```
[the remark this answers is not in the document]
akraitman: Since we have inventory hosts without ipmi interfaces I am not sure it's possible to combine those two tasks

roles/maas_nameserver/tasks/main.yml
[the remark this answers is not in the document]
djgalloway: I'm confused.  Those are hostnames.

roles/maas_nameserver/README.rst
djgalloway: I would suggest, instead, having a `default_domains` list in `defaults/main.yml` and have `maas` in it.

Then when deciding which domains to remove, combine the desired domains and the default domains.

akraitman: After checking why there is a default domain called "maas" it turns out that we can remove it so I can completely remove "allowed_domains" and I also don't think we need any default_domains variable in this case

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,98 @@
...
+These are role-specific variables, not intended for override:
+- `dns_domains`:
+  - `ceph`: Dynamically derived from the first Ceph host’s domain (e.g., `internal.ceph.tucson.com`).
+  - `ipmi`: Static IPMI domain (`ipmi.ceph.tucson.com`).
+
+## Secrets File
+The `maas_nameserver` role depends on the `secrets` role to load encrypted MAAS API credentials. The secrets file is expected at `{{ secrets_path }}/maas.yml`, where `secrets_path` defaults to `/etc/ansible/secrets`.
+
+### Example Secrets File
+Create and encrypt the file (e.g., `/etc/ansible/secrets/maas.yml`):
```
djgalloway, line 65: This should be `{{ secrets_path }}` with no mention of `/etc/ansible`.

akraitman:
> How feasible is it to test this against our existing ansible inventories?

I have tested it against a similar inventory like we have in sepia

roles/maas_nameserver/tasks/main.yml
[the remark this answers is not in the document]
akraitman: Task removed

roles/maas_nameserver/README.rst
akraitman: I have changed variable name to default_domains, we need to have this variable since the default domain is a dns domain that is used internally by maas, one example is when we deploy a machine it is used by maas for internal dns records so we need to exclude it from our ansible role

roles/maas_nameserver/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+  register: existing_resources
+  changed_when: false
+
+- name: Initialize DNS records list
+  ansible.builtin.set_fact:
+    dns_records: []
+
+- name: Define target hosts for DNS records
+  ansible.builtin.set_fact:
+    target_hosts: "{{ groups | dict2items | rejectattr('key', 'equalto', 'maas') | map(attribute='value') | flatten | unique | default([]) }}"
```
djgalloway, line 27: I think this should be a list defined in secrets.  Or at least mention that it can be overridden in secrets.

akraitman: I have added comments explaining its purpose and how to override it.

roles/maas_nameserver/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+maas_api_url: "http://localhost:5240/MAAS/api/2.0/"  # Default, overridden by secrets
+maas_profile: "admin"  # Default, overridden by secrets
+default_domains:
+  - "maas"
+
+# List of target hosts for DNS records, excluding the 'maas' group by default.
+# Can be overridden in secrets/maas.yml (e.g., target_hosts: ['host1.example.com', 'host2.example.com'])
+# or via extra vars (e.g., ansible-playbook -e "target_hosts=['host1.example.com']").
+# If not set, defaults to all hosts in inventory groups except 'maas'.
+target_hosts: []
```
djgalloway, line 11: Is this an ansible group that contains other ansible inventory groups?  Or is this saying "Only update DNS records for hosts in `target_hosts`?

roles/maas_nameserver/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+
+# Define target hosts for DNS records, excluding the 'maas' group.
+# This can be overridden in secrets/maas.yml (e.g., target_hosts: ['host1.example.com', 'host2.example.com'])
+# or via extra vars (e.g., -e "target_hosts=['host1.example.com']").
+- name: Define target hosts for DNS records
+  ansible.builtin.set_fact:
+    target_hosts: "{{ target_hosts | default(groups | dict2items | rejectattr('key', 'equalto', 'maas') | map(attribute='value') | flatten | unique | default([])) }}"
+  when: groups.keys() | length > 1
+
+- name: Build DNS records for all interfaces
```
djgalloway, line 33: Can you add a comment showing what this will look like after populated?

roles/maas_nameserver/vars/main.yml
djgalloway, line 1: dns_domains should be empty in this repo and overridden in the ansible inventory.

roles/maas_nameserver/defaults/main.yml
akraitman: target_hosts will just contain all hosts from inventory groups except 'maas', it's not an ansible_group

djgalloway: Is there a way to exclude certain ansible groups?  We may not want DNS records for all entries.

akraitman: Yes I can add this ability to this role, there is another thing I thought about that can also help us, its something I ran it to with ocp, for inventory hosts that need additional dns records like api.* there should be another inventory group used only for those additional dns records

roles/maas_nameserver/vars/main.yml
akraitman: Fixed and updated the changes in the readme file

roles/maas_nameserver/defaults/main.yml
akraitman: I have added the changes to be able to exclude certain ansible groups as described in the readme file

roles/maas_nameserver/vars/main.yml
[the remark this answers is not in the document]
akraitman: I removed it for now

roles/maas_nameserver/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+# or via extra vars (e.g., -e "target_hosts=['host1.example.com']").
+- name: Define target hosts for DNS records
+  ansible.builtin.set_fact:
+    target_hosts: "{{ target_hosts if target_hosts | length > 0 else (target_hosts_list | flatten | unique | default([])) }}"
+  when: groups.keys() | length > 0
+
+- name: Build DNS records for all interfaces
+  ansible.builtin.set_fact:
+    dns_records: "{{ dns_records + [{'name': item[0].split('.')[0], 'ip': interface_ip, 'type': 'A', 'domain': item[1].value}] }}"
+  loop: "{{ (target_hosts | default([])) | product(dns_domains | dict2items) | list }}"
```
dmick, line 50: I don't have the faintest idea what's going on here.  It looks as though this will make a list of two-element combinations of every item in target_hosts and every item in dns_domains.  How is that remotely useful?

I think at this point I'm going to have to stop reviewing until we can talk.  I'll save this review this far, but I'm not done.

dmick (changes requested):
Stopped midway through when I got to something I could not begin to explain.  We need to have a live session about this.

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,164 @@
...
+    tasks/
+      main.yml
+    meta/
+      main.yml
+    README.md
+```
+
+## Dependencies
+
+- **Secrets File**: A `maas.yml` file at `{{ secrets_path }}/maas.yml` provides MAAS API credentials (e.g., `maas_api_key`, `maas_api_url`). No separate secrets role is required; credentials are loaded via `include_vars`.
```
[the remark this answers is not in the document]
akraitman: I can place is under host_vars for maas.yml there is already an example in line 64

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Usage
+
+1. **Place the Role**
```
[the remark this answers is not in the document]
akraitman: Removed it

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+   Ensure the `maas_nameserver` role is in your `roles/` directory.
+
+2. **Prepare Inventory**
+
+   Define your `maas` and target host groups:
+
+   ```ini
+   [maas]
+   maas.internal.ceph.ibm.com ip=10.11.120.237
```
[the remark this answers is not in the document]
akraitman: I don't know of other hosts I am executing the role against the host running the maas api

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+   [snipeit]
```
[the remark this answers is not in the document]
akraitman: No, it's just an example

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,164 @@
...
+   snipe-it.internal.ceph.ibm.com ip=10.60.100.11
+
+   [machine]
```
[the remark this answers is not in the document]
akraitman: It's just an example, it doesn't mean anything, I could call it anything I just chosen some group name

roles/maas_nameserver/README.md
```diff
@@ -0,0 +1,164 @@
...
+     vlan104: "vlan104.internal.ceph.ibm.com"
+   ```
+
+4. **Set Up Secrets**
+
+   Create a secrets file at `{{ secrets_path }}/maas.yml`:
+
+   ```yaml
+   ---
+   maas_api_key: "XXXXXXXXXXXXXXXX"
```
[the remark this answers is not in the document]
akraitman: I agree that the url it's not a secret but I think it's a good practice to place the api_key, profile and url in the same place since we need all three to communicate with the maas api and yes 5240 is the default port

djgalloway:
I think you want to squash your commits.  Also, effort should be made to keep the actual commit message (the top line) as short as possible.  e.g., `maas_nameserver: Create role`.  All the bullet points under that are fine.

PR #781, opened 2025-07-01
Rocky 10 Support

djgalloway:


cobbler.yml
```diff
@@ -50,6 +50,7 @@
     - { role: cobbler_profile, distro_name: CentOS-8.stream-x86_64, tags: ['centos8.stream'] }
```
dmick, line 50: commit message "not defaults to" -> "now defaults to"

dmick (changes requested):
otherwise, sure

dmick:
OK with changes; what's up with test failures?

djgalloway:
> ok but those changes are a mess and should be cleaned up; also, I don't think we want to just remove lint warnings

I thought this is what you said you wanted 😆

dmick:
well, yes, but the last I looked Fernando had three commits and it looked done-ish.

PR #789, opened 2025-09-22
tools: cobbler-access password gen update

djgalloway:
- Work on Mac
- Make a letter/number the first character


dmick (approved):
why not.

PR #807, opened 2025-12-25
fix AnsibleFailedError: hwclock: Cannot access the Hardware Clock via…

guomuwen:
In virtual machines or containers, running the `smoke` test case can easily lead to failure due to errors like "['hwclock: Cannot access the Hardware Clock via any known method.', 'hwclock: Use the --verbose option to see the details of our search for an access method.']". This is because virtual machines or containers typically lack the necessary permissions to directly execute the `hwclock` command, and such errors are not usually serious and should not cause test case failures. In containers or virtual machines, `hwclock` errors should be ignored to allow the test case to continue execution.

fix https://tracker.ceph.com/issues/69280

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -31,11 +31,17 @@
   register: current_clocksource
 
 - name: Set the hardware clock
-  command: hwclock --systohc
+  command: hwclock --systohc --verbose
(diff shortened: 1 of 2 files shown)


djgalloway:
Closing as stale

PR #840, opened 2026-05-06
maas: Networking and Machine management tasks

djgalloway:


roles/maas/tasks/networking/fabric_vlans_read.yml
```diff
@@ -0,0 +1,20 @@
+---
```
dmick: maybe name this "fabric_vlans_read_from_maas" and/or add a comment that just says that "Read vlans for fabric fab_obj" and add a "returns" section stating "_vlans_raw_by_fabric".  Same for the other bits.  I think that'll increase the read speed a lot.

dmick (changes requested):
got to a file and had requests for commenting/naming conventions; if we can agree about it, I'll rereview once those are in place for the various files

dmick (approved):
This is pretty segregated, and while large, the usability and correctness will be easier to get from actual experience with it

PR #844, opened 2026-05-14
users: Change home dir path for teuthology host

djgalloway:
The home dirs live at the /tank/home mountpoint.  This change will default to /home unless the host we're running the role against is in the teuthology ansible group.

Fixes: https://tracker.ceph.com/issues/63562

### DIFF
--- a/roles/users/tasks/create_users.yml
+++ b/roles/users/tasks/create_users.yml
@@ -14,6 +14,7 @@
     shell: /bin/bash
     state: present
     append: yes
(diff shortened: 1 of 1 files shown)


batrick (approved):
Why not bind mount `/tank/home` to `/home`? Or a symlink.

anyway, if you prefer it this way +1

djgalloway:
> Why not bind mount `/tank/home` to `/home`? Or a symlink.

I'd have to hack up the ansible_managed and teuthology roles to keep the `cm`, `teuthology`, and `teuthworker` users' home dirs somewhere off the LRC so we can still SSH in in case the LRC is down.

PR #846, opened 2026-05-14
nameserver: Disable minimal-responses

djgalloway:
Just upgraded vpn-pub to el9.

BIND 9.16 (EL9) changed the default behavior for minimal-responses, causing authority and additional (glue) sections to be omitted from responses. Set minimal-responses no explicitly to restore consistent behavior across all nameservers regardless of BIND version.


dmick (approved):
<blinks>  okay

PR #850, opened 2026-06-01
Users role and nits

dmick:
Take care of some deprecation warnings, and completely redo the users role's
'filter_users' playbook.  It was ancient Jinja conditionals, and fragile wrt trying to
use strings to express objects.

It's probably easier to just compare the old code with the new code; diffs are impossible to read.

### DIFF
--- a/roles/users/tasks/filter_users.yml
+++ b/roles/users/tasks/filter_users.yml
@@ -1,39 +1,133 @@
 ---
-- name: Merge extra_admin_users into managed_admin_users
+# make a lookup of all users from name to dict
+- name: Build user lookup table
   set_fact:
-    # The following adds items from extra_admin_users to managed_admin_users, while
-    # fetching keys from the latter if they are not present in the former. It's as pretty
-    # as it can get without whitespace breaking the parser.
-    managed_admin_users:
-      "{% for new_admin in extra_admin_users -%}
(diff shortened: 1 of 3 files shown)


roles/users/tasks/filter_users.yml
```diff
@@ -1,39 +1,129 @@
...
   set_fact:
-    # The following rebuilds the managed_users list while omitting users already present
-    # in managed_admin_users
-    managed_users:
-      "[{% for lab_user in managed_users -%}
-      {% if not managed_admin_users|selectattr('name', 'equalto', lab_user.name)|list|length %}{{ lab_user }},{% endif %}
-      {%- endfor %}]"
-  when: extra_admin_users is defined and extra_admin_users|length > 0
-
-- name: Filter the managed_users list
```
djgalloway, line 25: Why `extra_admin_users is defined`?  Shouldn't it be when `managed_users` and `managed_admin_users` are defined and length not 0?

djgalloway:
Fine with the rest of it

roles/users/tasks/filter_users.yml
dmick: managed_admin at least.  I think managed_users is always defined.  but good catch.

djgalloway (approved):

PR #856, opened 2026-07-09
testnode: add Micron NVMe firmware update tasks

akraitman:


roles/testnode/tasks/main.yml
```diff
@@ -79,6 +79,9 @@
 - import_tasks: check-for-nvme.yml
   when: check_for_nvme == true
 
+- import_tasks: micron_nvme_firmware.yml
+  when: ansible_architecture == "x86_64"
```
djgalloway: Do the tasks not work on arm64?  We don't have any arm64 testnodes but may in the future.

djgalloway:
Looks good so far but a couple architecture changes I'd like:

1. Install nvme-cli and curl in the package installation tasks
2. Add some tasks in `roles/testnode/tasks/main.yml` that check the currently installed NVMe firmware version with the latest available in https://apt-mirror.sepia.ceph.com/lab-extras/micron-nvme-firmware/.  If the latest available matches what's currently installed, you can skip importing `micron_nvme_firmware.yml` entirely.
3. Don't make the NVMe firmware version comparison tasks fatal in case lab-extras is down or we're replacing the firmware binaries or something like that.

roles/testnode/tasks/main.yml
akraitman: I thought the Micron NVMe firmware flow would differ on ARM64, but the tasks work for both, so I dropped this when condition

roles/testnode/tasks/micron_nvme_firmware.yml
```diff
@@ -0,0 +1,197 @@
...
+- name: Reboot after Micron NVMe firmware update
+  reboot:
+    msg: "Rebooting after Micron NVMe firmware update"
+    reboot_timeout: 900
+  when:
+    - micron_nvme_firmware_update is defined
+    - micron_nvme_firmware_update.stdout is search("UPDATED:")
+  ignore_errors: true
+  failed_when: false
+
```
djgalloway, line 172: Need to add a `gather_facts:` or `setup:` after the reboot in case the drive numbering changes between reboots.

Found on a CentOS 9 stream trial node that the drive numbers changed after a reboot.  So the subsequent zap_disks tasks zap the wrong disk (the root disk).

akraitman: Good catch, I wasn't aware of it, I have added a setup task after the reboot

djgalloway (approved):

PR #862, opened 2026-08-12
alloy: ship journald/auditd to the LokiStack, take over node_exporter metrics from grafana-agent, remove it

djgalloway:
Adds a Grafana Alloy role + playbook for the lab's bare-metal and VM infra hosts. One Alloy per host runs two pipelines:

| Pipeline | Source | Destination |
| --- | --- | --- |
| **metrics** | node_exporter on `localhost:9100` | Mimir on the pok cluster |
| **logs** | systemd journal, `/var/log/audit/audit.log`, per-group allowlist | LokiStack on the pok cluster |

The metrics pipeline is what `roles/grafana_agent` did. **This role removes grafana-agent** (service, package, config, WAL, user) from every host it runs on; `roles/grafana_agent` and `grafana_agent.yml` are deleted and `alloy.yml` replaces them. grafana-agent static mode went EOL in Nov 2025 and shipping an EOL agent as the collector for a security audit trail was never the right tradeoff.

Companion to the `openshift-logging/` deployment in `ceph/sepia-openshift` (ceph/sepia-openshift#51).

## Status

- Logs: deployed and verified end to end on **soko05**. Queryable by host in the OpenShift console Logs view:
  ```
  {host="soko05.front.sepia.ceph.com", syslog_identifier=~"sshd|sudo|su|systemd-logind"}
  ```
- Metrics + grafana-agent removal: rendered configs for every pipeline combination pass `alloy validate`; `--check --diff` against soko05 (Ubuntu 24.04, both agents running) and adami08 (CentOS Stream 9, enforcing, grafana-agent only) shows exactly the intended plan. **Not yet applied for real anywhere.** The play is idempotent (`changed=0` on re-run) for the logs half; the remaining 204 hosts are not rolled out yet.

## Node Exporter Full for all hosts

`alloy.yml` covers 205 hosts. Today only **58** push `node_uname_info{job="node"}` to Mimir. The role installs the distro node_exporter wherever it's missing and scrapes it, so every host in scope lands in the **Node Exporter Full** dashboard (grafana.com 1860, `jenkins-grafana/dashboards/node-exporter-lab.yaml` in sepia-openshift, with the Mimir datasource selected). `count(node_uname_info{job="node"})` against Mimir is the rollout progress bar.

The label set grafana-agent produced is kept **verbatim** — `job="node"`, `instance="<ansible_fqdn>:9100"`, `nodename="<ansible_fqdn>"` — because that is what the dashboard's variables and the provisioned node-exporter alerts in sepia-openshift select on (they match literal values like `instance=~"chacra[123]:9100"`). Switching `instance` to `inventory_hostname` would orphan months of series and silently unhook those alerts, so it isn't done here even though `ansible_fqdn` is a mess (see below). The canonical `inventory_hostname` is carried as a `host` external label on **both** pipelines instead, so a host's metrics and logs join on one label. If we ever do move `instance`, it's one variable (`alloy_node_instance`) plus the alert expressions in sepia-openshift in the same change.

## Found by deploying, not by review

- **`ansible_fqdn` is unusable as the `host` label.** It reflects `/etc/hostname`, which is bare `soko05` on some hosts and a full FQDN on others — making `{host="..."}` queries unguessable across 205 machines. Uses `inventory_hostname`, which is canonical everywhere.
- **`main` was already broken for internal hosts' metrics.** `roles/grafana_agent` lost its `agent_mimir_url` default in 7cb459c, leaving every non-`public_facing` host with an undefined Mimir URL; hosts deployed before that still run with the old value. The in-cluster route is restored as the role default; `public_facing` still overrides it from the secrets repo group_vars under the same variable name, so nothing in secrets moves.
- **The Mimir route is served by the same pok ingress CA as the LokiStack gateway**, so TLS to Mimir is verified against the CA the role already ships instead of grafana-agent's `insecure_skip_verify`. `mimir.ceph.com` has a public cert and uses the system bundle. (When the ingress cert rotates, metrics *and* logs stop together — README covers it.)
- **Neither Mimir endpoint does auth today** — the in-cluster route and `mimir.ceph.com` both accept unauthenticated `remote_write` — and no deployed grafana-agent config ever carried `basic_auth` (the vaulted `mimir_password.yml` never defined both username and password). The role keeps the old semantics: basic auth is rendered only when both are defined. Worth fixing on the Mimir side separately; the role is ready for it.
- **EPEL 9 ships node_exporter as `node-exporter`** (`Provides: node_exporter`, aliased unit). If `:9100` is already taken and none of the known distro package names is installed — cephadm's node-exporter container on the LRC — the role leaves it alone and scrapes what's there rather than installing a second exporter to fight over the port.
- **The alloy RPM's `useradd -m -d /var/lib/alloy` hits the same SELinux denial grafana-agent's did**, so the `customuseradd` module moves into this role and loads before the package install.
- **Port collision.** Alloy and grafana-agent both default to `127.0.0.1:12345`. Alloy stays on 12346: Debian starts a service the moment its package installs, so on a host still running grafana-agent an Alloy on 12345 would crash-loop until the removal tasks and config handler caught up. grafana-agent is removed *before* Alloy is (re)started so both never push the same series at once.
- **The play must exclude `infra_compute`, `infra_storage`, `cnv`.** Those are the pok cluster's own RHCOS nodes, already covered in-cluster (cluster monitoring + ClusterLogForwarder), and they get pulled in implicitly through `[infrastructure:children]`.

## What the logs half collects

The systemd journal, `/var/log/audit/audit.log`, and an explicit per-group allowlist — deliberately **not** all of `/var/log`:

| Excluded | Why |
| --- | --- |
| `messages`, `secure`, `syslog`, `auth.log` | rsyslog copies of the journal — would double-ingest every line |
| `wtmp`, `btmp`, `lastlog` | binary; the login events are already in the journal as text |
| `*.gz`, `*.1` | rotated — would re-ingest all history on every logrotate |

## Secrets

Checked against `ceph-sepia-secrets`; no changes needed there:

- `loki_token.yml` (vault) — bearer token scoped to `create` on the `infrastructure` tenant only; cannot read logs back or touch the audit tenant.
- `mimir_password.yml` (vault) — already present; included for `mimir_username`/`mimir_password`, rendered only when both are set.
- `group_vars/public_facing.yml` — `agent_mimir_url` override to `mimir.ceph.com`, unchanged.
- The LokiStack/Mimir ingress CA ships with the role, since a CA certificate is public. Internal hostnames (LokiStack gateway, Mimir route) are role defaults, same as before.

`testnode`'s optional dependency becomes `run_alloy_role` (metrics only); nothing in secrets set the old `run_grafana_agent_role`.


akraitman:
Since Grafana Agent is already EOL and Alloy is its replacement, I think we should consider completing the migration rather than running both agents. We could move the existing metrics/node_exporter + Mimir remote_write configuration from the grafana_agent role into Alloy, validate that metrics are reaching Grafana correctly, and then disable/remove grafana-agent

roles/alloy/files/lokistack-ingress-ca.crt
akraitman, line 1: Will we need to renew this certificate on every machine running Alloy when they expire ? I think we should use Corporate/Public trusted CA instead so that we will only renew this trusted CA each time and then nothing extra required on the machines running Alloy

djgalloway: This pins the CA cert, not the service/router cert. The CA is good till December 2027. We don't have a corporate or public trusted CA in the lab but we could. That'd be a more in-depth follow-up PR in sepia-openshift.git. With that in place nothing needs pushing to the Alloy hosts at all.  cert-manager renews on the cluster and hosts trust it via the system bundle, same as mimir.ceph.com already works in this role.

akraitman (approved):

PR #868, opened 2026-08-14
fog-server: sideline stale EFI dirs on foreign ESPs after deploy

djgalloway:
Follow-up to #867 (this commit landed on that branch after it merged, so re-branched from main).

After a Deploy task, renames `EFI` → `EFI.stale` on ESPs of disks **other than the deploy target**, so a repurposed node's leftover install (e.g. a prior MAAS deployment) can't hijack UEFI boot into a GRUB whose BLS entries reference filesystems the deploy just destroyed (the trial-perf003/004 `VFS: unable to mount root fs` panic).

Scoped per review feedback: only runs for `trial*`/`smithi*` hosts (covers trial-perf), never touches the deploy target disk, and renames instead of deleting so it's reversible.

### DIFF
--- a/roles/fog-server/files/clean_foreign_boot.sh
+++ b/roles/fog-server/files/clean_foreign_boot.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Sourced by fog.postdownload inside the FOG (FOS) environment after a
+# Deploy task writes the image to disk.  Sidelines EFI bootloader files on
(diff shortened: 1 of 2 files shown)


djgalloway:
This wasn't necessary.  rEFInd was to blame.  Needed either EXIT or SANBOOT for the Exit Mode in FOG.

PR #869, opened 2026-08-17
wireguard: never let a config sync drop the interface's private key

djgalloway:
**Incident 2026-08-17:** the wireguard gateway (soko01) lost its private key on a config sync — `wg show wg0 public-key` returned `(none)`, all 127 peers' handshakes failed, lab VPN was down ~20 min until the key was re-set from the console.

**Root cause is in the role, not the trigger.** The rendered `wg.conf` deliberately has no `PrivateKey` (it's loaded by `PostUp = wg set %i private-key %i.key`). But `wg-quick strip` drops PostUp lines and `wg syncconf` *replaces* the interface config from its input — so the `sync wireguard config` handler (or the identical strip|syncconf done by hand) unsets the key every time it fires on a live interface. It only ever needed a changed conf to trigger.

Three independent fixes:
1. **template** — `PrivateKey` written into `wg.conf` explicitly as well as via PostUp; a raw strip|syncconf now keeps it.
2. **handler** — after strip, re-insert the key from the key file if absent; verify the candidate's public key matches the *live* interface's before syncing; verify the identity survived after; on any mismatch restore from the key file and fail. Failure leaves the running interface untouched.
3. **key task** — resolve the vaulted value via `set_fact` first (ansible-core 2.17 fails opaquely passing `!vault` content into `copy` under `no_log` — this is what sent us debugging in the first place), fail fast on an empty key, and notify the *sync* handler rather than a full restart (a key-file change must never drop every tunnel).

`--check --diff` against soko01 with these changes: the only diff is `+PrivateKey = …`.

Related: the onboarding service's post-merge Job no longer runs this role at all until this merges and a supervised first run proves it (see ceph/sepia-openshift#56).

### DIFF
--- a/roles/wireguard/handlers/main.yml
+++ b/roles/wireguard/handlers/main.yml
@@ -1,15 +1,46 @@
 ---
 # Apply config changes (e.g., peer additions/removals) to the running
-# interface without tearing down existing tunnels
-# Strip first and only sync if that succeeded; piping a failed strip
-# straight into syncconf would wipe all peers
+# interface without tearing down existing tunnels.
+#
+# INCIDENT 2026-08-17: `wg syncconf` REPLACES the interface's entire config
+# from its input, and `wg-quick strip` drops PostUp lines. Because the
+# rendered wg.conf deliberately carries no PrivateKey (it is loaded by
+# `PostUp = wg set %i private-key %i.key`), syncing the stripped conf left
+# the live interface with NO private key -- every peer's handshake failed
+# and the lab lost VPN until the key was re-set by hand from the console.
+#
+# So: strip, then re-append the private key from the key file, verify the
+# resulting config still yields the public key the interface has NOW, and
+# only then sync. Any failure leaves the running interface untouched.
 - name: sync wireguard config
   shell: |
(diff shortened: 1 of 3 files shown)


roles/wireguard/handlers/main.yml
```diff
@@ -1,15 +1,46 @@
...
+      conf="$stripped"
+    else
+      # Insert the key right after [Interface] so syncconf keeps it.
+      conf="$(awk -v key="$(cat "$keyfile")" \
+        '{print} /^\[Interface\]/ && !done {print "PrivateKey = " key; done=1}' <<< "$stripped")"
+    fi
+    want="$(wg show "$iface" public-key)"
+    have="$(grep '^PrivateKey' <<< "$conf" | head -1 | awk '{print $3}' | wg pubkey)"
+    if [ "$want" != "(none)" ] && [ "$want" != "$have" ]; then
+      echo "refusing to sync: config public key $have != live $want" >&2
```
dmick, line 31: verified that script output still happens even with no_log; no_log only suppresses Ansible's own logging

dmick (approved):

PR #872, opened 2026-08-19
testnode: baseurl creation is broken

dmick:
baseurl= overrides all previous baseurls, so with multiples we were only using the last one.  Change the template to output comma-separated baseurls, which should try them in order, which is what we want.


dmick:
https://jenkins.ceph.com/job/sepia-fog-images/2/ shows the problem; the appstream repo was accessing the last repo and failing to download repomd.xml, but the sepia mirror was just fine; running dnf -vv makecache showed it not even consulting the other three baseurls

djgalloway (approved):

PR #892, opened 2026-09-16
paddles: work with community.postgresql 5.0.0

djgalloway:
## What

teuthology's `requirements.yml` installs `community.postgresql` unpinned, so its ansible runs now get 5.0.0 and the paddles role fails at parse time:

```
ERROR! [DEPRECATED]: community.postgresql.postgresql_set has been removed. Use community.postgresql.postgresql_alter_system instead.
```

Reported in https://github.com/ceph/teuthology/pull/2256#issuecomment-5704434992. This fixes it on the ceph-cm-ansible side, so teuthology does not need to pin the collection.

### Collection compatibility

- `postgresql_set` (deprecated 4.0.0, removed 5.0.0) replaced with `postgresql_alter_system` (added 3.13.0, needs PostgreSQL >= 14). It reloads config itself, but `listen_addresses` needs a restart, so the restart-when-changed task stays.
- `db` alias of `login_db` was removed from `postgresql_info` (4.0.0) and `postgresql_user` (5.0.0). `login_db` works on every version, so use it.
- `postgresql_pg_hba` 3.x+ rejects a CIDR with host bits set. The role passed the docker_gwbridge *address* (`172.18.0.1/16`); 2.4.3 wrote it verbatim, 3.14.3 and 5.0.0 fail with "has host bits set". Use the interface's `network` fact instead. On an existing host this adds the new rule beside the old one; the role does not reload postgres, as before.
- FQCNs for the `community.postgresql` modules. The role now needs the collection at **>= 3.13.0**, documented in the README.

### Non-containerized path fixes

Found while exercising the role on Ubuntu 24.04. Production paddles is containerized, so none of these change what runs there:

- PostgreSQL >= 15 no longer grants CREATE on schema `public` to every role, so `pecan populate` failed. Grant the paddles DB user ALL on `public`, only alongside DB creation so existing deployments are untouched.
- Python >= 3.12 venvs ship no setuptools, and setuptools >= 82 dropped `pkg_resources`, which pecan 1.3.3 (pinned by paddles) still imports. Install `setuptools<82` into the venv; drop once paddles moves to pecan >= 1.5.
- `source virtualenv/bin/activate` is a bashism under `/bin/sh` (dash), so `alembic stamp head` never ran. Put the venv bin dir on `PATH` instead (alembic's env.py shells out to `pecan`, so calling the binary directly is not enough).
- Ensure supervisord is enabled and started before `supervisorctl update`.

## Testing

All against throwaway Ubuntu 24.04 systemd containers in local podman, never against production:

| Run | Result |
|---|---|
| Full role from scratch, collection 5.0.0 | pass, paddles answers HTTP 200 via nginx |
| Full role from scratch, collection 3.14.3 | pass |
| Rerun on both | idempotent (only the users role's `git ls-remote` tasks report changed) |
| Containerized-path `postgresql_info` / `postgresql_pg_hba` tasks, 5.0.0 and 3.14.3 | pass with the network fact, fail with the old address form |
| Collection 2.4.3 | fails at parse (module not found), hence the README requirement |

## Reviewer notes

- Production PostgreSQL must be >= 14 for `postgresql_alter_system`. Not verified from this session.
- Whoever runs `paddles.yml` needs `community.postgresql >= 3.13.0` locally (`ansible-galaxy collection install 'community.postgresql:>=3.13.0' --upgrade`).


kshtsk (approved):
Looks fine as far as it passes `ansible-playbook --check`

djgalloway:
> Looks fine as far as it passes `ansible-playbook --check`

Tested end to end on a fresh Ubuntu 24.04 container then ran again to test idempotency.

kshtsk (approved):

akraitman (approved):

PR #893, opened 2026-09-17
reverse_proxy: manage soko01's pulpito vhost; add GET-only paddles proxy

djgalloway:
soko01's nginx vhosts have been hand-edited since the Anubis rollout and were not tracked anywhere. This adds a `reverse_proxy` role (and `reverse_proxy.yml`) that brings `pulpito.conf` under git, rendered from a template that matches the live file byte for byte, and adds one new location on pulpito.ceph.com:

```
location ~* ^/_paddles/runs/   ->   https://paddles.front.sepia.ceph.com/runs/...
```

- `limit_except GET` and a source-IP allowlist (paddles has no authentication of its own, so the method limit is what keeps the write endpoints private)
- 60s cache in the existing `paddles_cache` zone, bypasses Anubis like the neighbouring suites/branches/machine_types bypass
- the rest of `/_paddles/` and everything else on the vhost is unchanged

**Why:** an external CI team asked for a WireGuard peer for a Jenkins server in IBM Cloud so a dashboard could read nvmeof run/job results. A soko01 peer gets unrestricted forwarding into every lab subnet including IPMI, and a public Jenkins with password auth is not a host to hand that to. Their need is two read-only paddles endpoints, which were not reachable from the internet at all before this, so expose exactly those to their address instead.

**Allowlist as data:** role default `reverse_proxy_paddles_readonly_allow` only carries the lab's public /24. Third-party addresses go in `reverse_proxy_paddles_readonly_allow_extra` in the secrets repo host_vars so they stay out of this public repo (companion PR in ceph-sepia-secrets adds the `[reverse_proxy]` group and soko01's host_vars).

**Deployment:** already applied by hand on soko01 on 2026-09-17 and verified (allowed GETs 200, non-GET 403, other sources 403, existing paths unchanged). No play was run against the VPN gateway. Once the secrets PR is merged, `ansible-playbook reverse_proxy.yml --check --diff` should be a no-op.

Assisted-by: Claude Fable 5.1
Signed-off-by: David Galloway <david.galloway@ibm.com>


djgalloway:
Superseded: the same GET-only paddles proxy is being added to the pulpito template in #879, which already models soko01's reverse proxy stack under public_facing (and secrets #1227 already put soko01 in that group). This role would have duplicated it and reused the reverse_proxy_sites name with a different shape.

PR #894, opened 2026-09-18
public_facing: throttle heavy mirrors on download.ceph.com

djgalloway:
## Why

download.ceph.com runs on an OVH instance with capped public bandwidth, and its data volume is capped at 250 IOPS. In Sep 2026 a package-mirror crawler on GoDaddy's network held ~560 HTTPS connections and pulled ~2 TB/day. Egress sat pinned at the cap for two weeks and every other client, including our own builders, got ~50 KB/s.

The operator reached out and it was unintentional. So instead of blocking them, allow them but limit them, and make that repeatable for the next heavy mirror.

## What

Two layers, because they solve different problems:

- **`/etc/nginx/conf.d/throttle.conf`**: per-source-IP connection, request-rate and per-connection rate limits for the CIDRs in `download_throttled_subnets`. Hundreds of slow connections exhaust the volume's IOPS even when the pipe isn't full, so a bandwidth cap alone isn't enough. The limit keys are empty for everyone else, so no other client is affected. Every directive is at `http` level, so the vhost (which this role does not manage) needs no changes.
- **`egress-shaper.service`**: an HTB tree built with `tc`. One class for rsyncd (tcp/873), one aggregate class for `download_throttled_subnets`, unshaped default. With 26 source IPs, the nginx limits alone would still permit ~800 Mbit/s in aggregate.

All limits are `download_*` variables in `defaults/main.yml`:

| Variable | Default |
|---|---|
| `download_rsync_egress_limit` | `20mbit` |
| `download_throttled_subnets` | `64.202.160.0/24`, `64.202.174.0/24` |
| `download_throttled_subnets_egress_limit` | `50mbit` |
| `download_throttle_conn_per_ip` | `2` |
| `download_throttle_req_per_sec` / `_burst` | `5` / `20` |
| `download_throttle_rate_per_conn` | `2m` |
| `download_public_iface` | interface of the default IPv4 route |

Also adds a download.ceph.com section to the README, covering these and the existing hourly `make_timestamp` cron.

## Testing

- Templates rendered with the default values and installed verbatim on the replacement download.ceph.com instance (b3-16, Ubuntu 24.04): `nginx -t` passes, `egress-shaper.service` is active, `tc class show` lists the 20 Mbit and 50 Mbit classes.
- Rapid requests from a client outside the listed subnets all return 200, so the empty-key exemption works.
- Not yet run through `ansible-playbook public_facing.yml --limit download.ceph.com`; the host currently carries hand-installed copies of the rendered files, so the first run should be a no-op.


roles/public_facing/templates/download.ceph.com/egress-shaper.sh.j2
```diff
@@ -0,0 +1,31 @@
...
+# Egress shaping for download.ceph.com.  OVH caps the instance's public
+# bandwidth, so a single heavy mirror can starve everyone else.
+#   1:10  rsyncd (tcp/873)                 {{ download_rsync_egress_limit }}
+#   1:30  download_throttled_subnets       {{ download_throttled_subnets_egress_limit }} (aggregate)
+#   1:20  everything else                  unshaped
+DEV={{ download_public_iface }}
+
+tc qdisc del dev $DEV root 2>/dev/null || true
+tc qdisc add dev $DEV root handle 1: htb default 20
+tc class add dev $DEV parent 1:  classid 1:1  htb rate 10gbit ceil 10gbit
```
akraitman: I would recommend making the root HTB bandwidth configurable instead of hard-coding it to 10gbit.

For example, we could define something like:

download_egress_rate: 1gbit

and use that value for the root HTB class.

The idea is that the root HTB rate should represent the actual public bandwidth available on the OVH instance. If OVH limits the instance to 1 Gbit but we configure the root class as 10 Gbit, tc is working with a link capacity that doesn't reflect the real bottleneck.

The existing 20mbit rsync and 50mbit throttled-subnet limits would still work, but having the root rate match the actual OVH bandwidth would make the overall shaping configuration more accurate and easier to adjust if the instance or bandwidth limit changes later.

roles/public_facing/templates/download.ceph.com/egress-shaper.sh.j2
akraitman: Would it make sense to add set -e (or set -eu) to egress-shaper.sh? That way, if one of the tc commands fails, systemd will report the service as failed instead of potentially leaving us with a partially applied configuration

